PR #10, opened 2025-07-02
README.md initial draft

jorge-campo:
@cheny0 , let me know if there's any section I'm missing or if you have suggestions to improve the README.

Some notes:

- We will include a simple contributor's guide in the README and link to the Style Guide for detailed instructions.
- I created a table that lists the different project phases and their corresponding statuses (traffic lights).
- Each week, we'll feel out what's new for the week, with links to the corresponding discussion or issues.
- Some parts of the README are commented. We'll make them visible when things are ready.
- I will ask Francesca for a nice illustration to use as a header.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,19 +47,19 @@ All documentation lives in **one monorepo** (this repo); each project has its ow
 
 ## Project status
 
-| #  | Phase                              | State |
-|----|------------------------------------| :---: |
-| 1  | Project plan and kick-off          | 🟡    |
-| 2  | GitBook and domain setup           | ⚪    |
-| 3  | Repo scaffold                      | 🟡    |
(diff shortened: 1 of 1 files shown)


jorge-campo:
> It looks good to me. I just have one concern: what if the weekly update becomes a lengthy list? It can be stressful to see a big README. How about moving the update to another file?

That's the idea: to keep only the last one to two weeks in the README file and move the old records to another file. This was the  intention with these lines: https://github.com/status-im/ift-docs/pull/10/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R81-R83

cheny0 (approved):

PR #18, opened 2025-07-09
Docs Standards initial files and folders layout

jorge-campo:
@cheny0 , this is the initial "Docs Standards" structure, including the empty Markdown files for the concept types description (quickstarts, concepts, etc.)

It isn't final and doesn't include all the sections that we may need. Let's add/remove things as we move forward with other parts of the project.


cheny0 (approved):
LGTM!

Issue #19, opened 2025-07-09
Content attributions in project documentation


jorge-campo:
We use GitHub Docs examples in many of our examples and templates.

We must properly attribute these projects in the `ATTRIBUTIONS.md` file and at the end of every file where we copy large blocks of text using:

> Adapted from GitHub Docs (CC BY 4.0)

PR #20, opened 2025-07-09
Add ATTRIBUTIONS.md with GitHub Docs attribution

jorge-campo:
We use GitHub Docs examples in many of our examples and templates.

We must properly attribute these projects in the `ATTRIBUTIONS.md` file and at the end of every file where we copy large blocks of text using:

> Adapted from GitHub Docs (CC BY 4.0)

`closes: #19`


cheny0 (approved):
LGTM!

PR #35, opened 2025-07-24
Create editorconfig file

cheny0:
I combined the file from the status-website repo (https://github.com/status-im/status-website/blob/main/.editorconfig) and some recommendations from ChatGPT

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,14 +9,6 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = false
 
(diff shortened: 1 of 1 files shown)


jorge-campo:
> @jorge-campo, because indent size of 2 is used for status website repo and we've been using it for Status docs. ChatGPT also mentioned that many Markdown editors and static site generators (like Docusaurus, GitBook, and others) default to 2-space indentation.

Got it! If you think this works best for us, let's do it.

PR #123, opened 2026-01-05
Update admonition writing rules

cheny0:


manual/3-validate-the-design/writing-rules/style-the-content/admonitions.md
jorge-campo: Let's use the "callout" term instead of "admonition".

For the audience we target (developers, mainly), "admonition" is not a well-known term. It also seems that this term is used less and less. For example, in the Microsoft Style Guide, "admonition" [returns no results](https://learn.microsoft.com/en-us/search/?scope=Style%20Guide&category=Documentation&terms=admonition); [callout does](https://learn.microsoft.com/en-us/search/?scope=Style%20Guide&category=Documentation&terms=callout).

PR #126, opened 2026-01-12
Updates to procedure, concept, and quickstart templates

jorge-campo:
This PR includes updates to the following files:

- `concept-template.md` (and the corresponding JSON template).
- `procedure-template-flat.md`
- `procedure-template-sectioned.md`
- `quickstart.md`
- The JSON templates corresponding to the Markdown files above
- `transform.py` script in charge of the Markdown to JSON transformation
- `md2json-prompt-templates.md` prompt for the transformation process

The files include multiple changes to improve LLM readability (for example, examples are no longer represented in Markdown tables in the JSON files and now include IDs) and to better control edge cases.

> [!NOTE]
>
> At this point, without being able to test these JSON files for LLM-based authoring, it's not worth updating these templates further. The Markdown templates are now all we need to write standardized content without LLM support.

### DIFF
--- a/manual/2-populate-the-structure/templates/quickstart/quickstart-template.json
+++ b/manual/2-populate-the-structure/templates/quickstart/quickstart-template.json
@@ -690,6 +690,12 @@
           "group": "STRUCT",
           "description": "Do not use H4-H6 headings.",
           "examples": []
+        },
+        {
+          "id": "PROC-BEHAV-FORBID-NO-FURTHER-READING",
+          "group": "BEHAV",
+          "description": "Do not include a \"Further reading\" section or links to other related topics at the end of the document.",
+          "examples": []
         }
       ],
       "examples": []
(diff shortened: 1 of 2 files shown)


manual/2-populate-the-structure/templates/procedure/procedure-template-flat.json
```diff
@@ -912,7 +927,7 @@
         {
           "id": "PROC-STRUCT-FORBID-H4-H6",
           "group": "FORBID",
```
cheny0:
```suggestion
          "group": "STRUCT",
```

jorge-campo: Fixed, thanks.
One note, @cheny0: we can't make changes to the JSON files. Well, we can but we *shouldn't*. These files are made programmatically by the `transform.py` script and Markdown prompt. If somethings is wrong in the JSON file, we must fix it in the Markdown file otherwise, the automatic MD2JSON transformation is useless. I think you already knew that.

PR #211, opened 2026-03-25
Zone SDK Tutorial

kashepavadan:
Tutorial for using the Zone SDK to build a simple custom Zone, with the sql password manager Zone in the logos-blockchain sql-zone branch as the example to follow.


docs/blockchain/zone-sdk-tutorial.md
```diff
@@ -0,0 +1,544 @@
...
+```
+
+## Conclusion
+### Using the Password Manager
+After you add the code from the tutorial, your password manager should be ready to use. Instructions on building and interacting with the sequencer and indexer applications can be found within the `testnet/sqlite-zone-demo` folder, in the `README.md` file.
+
+### More Zone Possibilities
+This tutorial illustrated the basic functionality of the Zone SDK to build a simple appchain Zone. However, this is far from the only possibility for custom Zones on Logos. You could use the SDK to build traditional ZK or optimistic rollups, customised high transaction throughput appchains, or even applications with functionality compartmentalised across several Zones.
+
+Despite their autonomy and customisability, Logos Zones support several key interoperability features. Bridging Layer 1 (Bedrock) tokens into Zones, on-chain message passing between Zones, as well as complex cross-Zone transaction coordination are all supported by the Logos Blockchain. See [**The Secret to Sovereign Zone Interoperability on Logos**](https://press.logos.co/article/sovereign-zone-interoperability) for more details.
```
davidrusu: We will be calling the Bedrock "Logos Blockchain", Shouldn't be referring to bedrock publicly

kashepavadan: Will this change in the specs as well?

weboko (commented):
Dogfooded on Ubuntu 24.04.4 LTS x86_64 with Rust 1.94.1. Cloned `logos-sql-zone`, switched to the `tutorial` branch, and followed the tutorial by adding all the code examples to the skeleton files.

Two issues blocked a successful build — see inline comments. After fixing them, `cargo build` succeeded with no errors.

The tutorial logic and code examples are otherwise correct once these gaps are addressed.

docs/blockchain/zone-sdk-tutorial.md
```diff
@@ -0,0 +1,660 @@
...
+The complete demo is available in the `master` branch.
+
+### Access to Logos Blockchain Node
+It's also important for your sequencer and indexers to have access to a Logos Blockchain Node. See the [documentation](../quickstart-guide-for-the-logos-blockchain-node.md) for instructions on how to start your own node and connect to the Testnet.
+
+## Sequencer
+### Define Your Zone Environment
+The first step to creating a Logos Zone is defining the Zone state - what application(s) it will host, how the state is updated, and in what form updates are posted on-chain.
+
+#### Example
```
weboko: **Missing step: git submodule init.** The `logos-sql-zone` repo uses `logos-blockchain` as a git submodule. After running `git clone` and `git checkout tutorial`, the `logos-blockchain/` directory is empty, so `cargo build` fails immediately with:
```
error: failed to load manifest for dependency `logos-blockchain-core`
Caused by: No such file or directory
```

A submodule init step is required after the clone:

```suggestion
```bash
git clone https://github.com/logos-blockchain/logos-sql-zone.git
cd logos-sql-zone
git submodule update --init --recursive
```
```

File: docs/blockchain/zone-sdk-tutorial.md [hunk 3]
```diff
@@ -0,0 +1,660 @@
...
+    pub async fn run(self) {
+
+        ...
+
+        loop {
+            let Some(event) = sequencer.next_event().await else { continue; };
+            handle_event(event, &handle, &mut state, &checkpoint_path).await;
+        }
+    }
+}
```
weboko: **Bug: `ZoneIndexer` missing generic type parameter.** The tutorial shows:
```rust
pub struct Indexer {
    zone_indexer: ZoneIndexer,
    db_path: String,
}
```
But `ZoneIndexer<Node>` requires a generic argument. This fails to compile:
```
error[E0107]: struct takes 1 generic argument but 0 generic arguments were supplied
```

Suggested fix:
```suggestion
pub struct Indexer {
    zone_indexer: ZoneIndexer<NodeHttpClient>,
    db_path: String,
}
```
`NodeHttpClient` is already imported via `use logos_blockchain_zone_sdk::adapter::NodeHttpClient;` in the tutorial’s imports.

docs/build-an-app/tutorials/build-dapp-with-logos-zone-sdk.md
```diff
@@ -0,0 +1,658 @@
...
+
+1. In `indexer/src/indexer.rs`, add the `Indexer` struct definition:
+
+   ```rust
+   // Indexer struct
+   //
+   // zone_indexer: SDK indexer
+   // db_path: Path to password database file
+   pub struct Indexer {
+       zone_indexer: ZoneIndexer,
```
pradovic: I think the indexer is missing generic param: `﻿﻿zone_indexer: ZoneIndexer<NodeHttpClient>`

docs/build-an-app/tutorials/build-dapp-with-logos-zone-sdk.md
```diff
@@ -0,0 +1,658 @@
...
+           })?;
+           let channel_id = parse_channel_id(channel_id_str.trim())?;
+   
+           info!("Channel ID: {}", hex::encode(channel_id.as_ref()));
+   
+           // New ZoneIndexer
+           let node = NodeHttpClient::new(CommonHttpClient::new(basic_auth), node_url);
+           let zone_indexer = ZoneIndexer::new(channel_id, node);
+   
+           Ok(Self { zone_indexer, db_path.to_owned() })
```
pradovic:
```suggestion
           Ok(Self { zone_indexer, db_path: db_path.to_owned() }
```

docs/build-an-app/tutorials/build-dapp-with-logos-zone-sdk.md
```diff
@@ -0,0 +1,658 @@
...
+       debug!("Processing batch of {} queries", count);
+   
+       // Publish batch of statements
+       let sql_bytes = pending.join("\n").into_bytes();
+       let inscription = Inscription::try_from(sql_bytes)
+           .map_err(|e| SequencerError::InscriptionTooLarge(e.to_string()))?;
+       if let Err(e) = client.publish(inscription).await {
+           error!("failed to publish batch: {e}");
+       } else {
+           info!("Submitted batch of {} statement(s)", count);
```
pradovic, line 284: Maybe we could call on_published method from state here and give it the id for bookkeeping?

pradovic (commented):
Looks good overall! I left a few comments to consider.

docs/build-an-app/tutorials/build-dapp-with-logos-zone-sdk.md
```diff
@@ -0,0 +1,658 @@
...
+           }
+           Event::BlocksProcessed { checkpoint, channel_update, finalized } => {
+               state.on_adopted(&channel_update.adopted);
+               for orphan in &channel_update.orphaned {
+                   let OrphanedTx::Inscription(info) = orphan else { continue };
+                   state.on_orphaned(&info.this_msg);
+   
+                   // Republish orphaned messages
+                   debug!(msg_id = %hex::encode(info.this_msg.as_ref()), "Auto-republishing orphan");
+                   if let Err(e) = sequencer.handle().publish(info.payload.clone()) {
```
pradovic: In a multi-sequencer Zone, a competing inscription can orphan yours while a logically-equivalent update is adopted under a different message ID — consumers should dedup by an app-level identifier (e.g. a uuid in the payload) against channel_update.adopted before republishing." The DECENTRALIZED_SEQUENCING.md doc and BalanceAwarePolicy are the right reference.

If the example is for single sequnecer, then I think ChannelUpdate can be completely ignored and you can just save the checkpoint/finalized. In that case no need to track published at all, as we don't need to compare payloads to dedup.

pradovic:
@kashepavadan Is this a single sequencer example? If it is I think it can be made more simple with less state tracking and without a resubmittion policy.

Even more if the indexer is used to observe finalized blocks, then the only thing that is needed to persist might be checkpoint.

kashepavadan:
@pradovic Yes, this is a single sequencer scenario. Don't we still need resubmission in case of reorgs? How would you change it?

pradovic:
> @pradovic Yes, this is a single sequencer scenario. Don't we still need resubmission in case of reorgs? How would you change it?

No, in single sequnecer scenario reorg does not change parent lineage, so auto resubmittion will take care of this. In single scenario you can ignore the ChannelUpdate event, only save checkpoint on each BlocksProcessed event

docs/build-an-app/tutorials/build-dapp-with-logos-zone-sdk.md
```diff
@@ -0,0 +1,622 @@
...
+           })?;
+           let channel_id = parse_channel_id(channel_id_str.trim())?;
+   
+           info!("Channel ID: {}", hex::encode(channel_id.as_ref()));
+   
+           // New ZoneIndexer
+           let node = NodeHttpClient::new(CommonHttpClient::new(basic_auth), node_url);
+           let zone_indexer = ZoneIndexer::new(channel_id, node);
+   
+           Ok(Self { zone_indexer, db_path.to_owned() })
```
weboko: This struct initializer is invalid Rust — the reader copies it verbatim into `indexer/src/indexer.rs` and it won't compile (`db_path.to_owned()` is a bare expression where a field initializer is expected). It needs the `db_path:` field name. Confirmed against the complete implementation on the repo's `master` branch, which has `Ok(Self { zone_indexer, db_path: db_path.to_owned() })`.

```suggestion
           Ok(Self { zone_indexer, db_path: db_path.to_owned() })
```

weboko (changes requested):
Reviewed the added Zone SDK tutorial. One copy-paste compile bug in the Indexer `new()` snippet (inline). The rest of the structural claims (InMemoryZoneState in `common/src/state.rs`, the `ZoneState` trait, `handle_event`) match the `logos-sql-zone` `master` solution branch. Note: I did not run the full tutorial end-to-end (it needs a running Logos Blockchain node + the sql-zone submodules); this bug was found by comparing the snippets against the solution branch.

docs/build-an-app/tutorials/build-dapp-with-logos-zone-sdk.md
weboko: Heads-up: this suggestion drops the closing `)` (`...to_owned() }` — `Ok(` is left unclosed), so applying it as-is still won't compile. The complete fix is `Ok(Self { zone_indexer, db_path: db_path.to_owned() })` (see my suggestion on this line).

docs/build-an-app/tutorials/build-dapp-with-logos-zone-sdk.md
```diff
@@ -0,0 +1,622 @@
...
+
+1. In `indexer/src/indexer.rs`, add the `Indexer` struct definition:
+
+   ```rust
+   // Indexer struct
+   //
+   // zone_indexer: SDK indexer
+   // db_path: Path to password database file
+   pub struct Indexer {
+       zone_indexer: ZoneIndexer,
```
weboko: `ZoneIndexer<Node>` is generic with no default type, so a bare `ZoneIndexer` here fails to compile:

```
error[E0107]: missing generics for struct `ZoneIndexer`
  --> struct takes 1 generic argument but 0 generic arguments were supplied
```

In `Indexer::new()` below, `zone_indexer = ZoneIndexer::new(channel_id, node)` with `node: NodeHttpClient`, so the concrete type is `ZoneIndexer<NodeHttpClient>` — matching how the sequencer struct is already declared (`sequencer: ZoneSequencer<NodeHttpClient>`).

```suggestion
       zone_indexer: ZoneIndexer<NodeHttpClient>,
```

weboko (commented):
Re-reviewed the latest commit (`b97fae5`), which applied my `Ok(Self { zone_indexer, db_path: db_path.to_owned() })` fix — that one's resolved, thanks.

While re-checking the Indexer section against the SDK I found one compile bug still present in the current head: the `Indexer` struct field types `zone_indexer` as a bare `ZoneIndexer`, but `ZoneIndexer<Node>` is generic with no default (verified in `zone-sdk/src/indexer.rs`: `pub struct ZoneIndexer<Node>`). Note the sequencer struct a few sections up already gets this right (`sequencer: ZoneSequencer<NodeHttpClient>`). Detail + suggested fix inline.

PR #231, opened 2026-04-23
Fix run command: use LogosBasecamp instead of logos-app-poc

veriqontrust:
Updated the build/run guide to use the correct binary name (`LogosBasecamp`) instead of `logos-app-poc`.  
Added a note about legacy builds to avoid confusion for newcomers.


cheny0:
Thanks for reviewing the docs, @veriqontrust ! This document was created under time constraints, and we’re currently working on a new version to replace it.

veriqontrust:
> Thanks for reviewing the docs, @veriqontrust ! This document was created under time constraints, and we’re currently working on a new version to replace it.

My pleasure , just trying my best to atleast be of some use.

PR #233, opened 2026-04-24
(#232) Update the issue template

### LINKED ISSUE #232: Update the issue template
Doc packets for UIs should include
- explanation of what you can do with the app
- flow of the most interesting action with no screenshot/videos, just button labels "click 'new group';" to start group chat, etc

cheny0:


cheny0:
@fryorcraken, I added only one line here since the other requirements are the same as for doc packets that don’t involve UI elements

Issue #234, opened 2026-04-24
[Journey] Store and retrieve a file using the Simple Filesharing App


gmega:
### What the user achieves

The user has managed to share and download file(s) from the Logos storage network

### Why it matters

Logos storage is a filesharing app, this enables people to actually share files.

### Key components

The initial bootstrap will rely on public (testnet) bootstrap nodes. Those are hardcoded into the app.
When sharing, nodes will initially connect to other nodes [over the DHT protocol](https://lip.logos.co/storage/raw/dht.html) to advertise file descriptors.
When downloading, nodes will run lookup queries using [the DHT protocol](https://lip.logos.co/storage/raw/dht.html), whereas downloads will happen using [the block protocol](https://lip.logos.co/storage/raw/codex-block-exchange.html) over libp2p.

### Repository

https://github.com/logos-co/logos-storage-ui @ bfa0ae6196d629e6bbf8f028a3fc8ad58f7645dc

### Runtime target

testnet v0.1

### Prerequisites

- OS: I've tested this on Ubuntu 24.04.4 LTS, but it should work on MacOS (Tahoe, earlier releases too) and pretty much any marginally recent 
Linux distro.
- Hardware requirements: 
  - consumer-grade CPU and 4GB RAM (should run on an RPi);
  - as much disk space as there are files you want to share/download.

### Commands and expected outputs

See the [Application Manual](https://github.com/logos-co/logos-storage-ui/blob/master/README.md#using-the-application).

### Success command

You should see the files you download in the filesystem, so `ls` should do. :-) 

### Expected result

Shared files should be available to other users, downloaded files should be accessible in the local filesystem.

### Configuration details

Those are addressed in the app's README file: https://github.com/logos-co/logos-storage-ui/blob/master/README.md

### Failure modes and limits

Again, addressed in the README file: https://github.com/logos-co/logos-storage-ui/blob/master/README.md#troubleshooting

### GitHub handle

gmega

### Discord handle

giuliano.mega

### Existing docs or specs

The app's README contains all the relevant information: https://github.com/logos-co/logos-storage-ui/blob/master/README.md

### Hardware requirements

Already described above.

### Estimated time to complete

10 minutes in case there are no snags with network config

### Security notes

Nothing noteworthy.

PR #298, opened 2026-05-29
Store and Retrieve a File using the Storage UI

kashepavadan:
Related to: https://github.com/logos-co/logos-docs/issues/234

Development info from doc packet has been removed as this is a beginner tutorial


docs/storage/get-started/set-up-and-use-logos-storage-ui.md
```diff
@@ -0,0 +1,190 @@
...
+
+The application is built using Nix flakes. The output includes the storage UI plugin and supporting binaries.
+
+{% hint style="info" %}
+
+If Nix flakes are not enabled globally, add `experimental-features = nix-command flakes` to `~/.config/nix/nix.conf`.
+
+{% endhint %} 
+
+1. Clone the repository and enter the project directory.
```
danisharora099: The doc never says which repo to clone, and the only repo linked above is `logos-storage-module` — but that builds the module, not the UI. The UI app lives in `logos-co/logos-storage-ui`. Making it explicit:

````suggestion
1. Clone the [`logos-storage-ui`](https://github.com/logos-co/logos-storage-ui) repository and enter the project directory:

   ```bash
   git clone --recurse-submodules https://github.com/logos-co/logos-storage-ui.git
   cd logos-storage-ui
   ```
````

File: docs/storage/get-started/set-up-and-use-logos-storage-ui.md [hunk 2]
```diff
@@ -0,0 +1,190 @@
...
+   │   ├── logos-storage-ui-app  # Standalone Qt application
+   │   ├── logos_host           # Logos host executable (for plugins)
+   │   └── logoscore            # Logos core executable
+   ├── lib/
+   │   ├── liblogos_core.dylib  # Logos core library
+   │   └── liblogos_sdk.dylib   # Logos SDK library
+   ├── modules/
+   │   ├── capability_module_plugin.dylib
+   │   └── storage_module_plugin.dylib
+   └── storage_ui.dylib            # Qt plugin (loaded by app)
```
danisharora099: This tree doesn't match what `nix build` produces — there's no `bin/`, `logoscore`, or `modules/`. The default output is just the plugin under `result/lib/`:

```suggestion
   result/
   └── lib/
       ├── storage_ui_plugin.dylib           # Qt plugin (loaded by the app)
       └── storage_ui_replica_factory.dylib
```

docs/storage/get-started/set-up-and-use-logos-storage-ui.md
```diff
@@ -0,0 +1,190 @@
...
+   ```
+
+1. Enter development shell (Optional)
+
+   ```bash
+   # Enter development shell with all dependencies
+   nix develop
```
danisharora099, line 74: The bash code block opened just above is never closed, so the fence runs into the next step and the rest of the page renders wrong. Closing it here:

````suggestion
   nix develop
   ```
````

File: docs/storage/get-started/set-up-and-use-logos-storage-ui.md [hunk 4]
```diff
@@ -0,0 +1,190 @@
+---
+title: Set up and use the Logos Storage UI
+doc_type: procedure
+product: storage
+topics: []
+steps_layout: sectioned
+authors:
```
danisharora099: `authors:` is empty (and `topics:` is `[]`). Mind filling these in — every other journey lists its authors.

docs/storage/get-started/set-up-and-use-logos-storage-ui.md
```diff
@@ -0,0 +1,190 @@
...
+
+   ```bash
+   nix run
+   ```
+
+   - To override a dependency with a local version, use `--override-input`. For example:
+     ```bash
+     nix run --override-input storage_module/logos-storage git+file:///somewhere/logos-storage-nim?submodules=1
+     ```
+   - To run from `result/` directly after `nix build`, use `./result/bin/logos-storage-ui-app`.
```
danisharora099: There's no `result/bin/` from the default `nix build` (it outputs the module under `result/lib/`), so this path doesn't exist. `nix run` is the way — which the doc already says up in Step 1.2.

danisharora099 (commented):
Heads up — this one needs a couple of fixes before it's followable end to end. The build and the app itself are fine (I got `nix build` producing the plugin and `nix run` bringing the app up — it loads storage_module, spawns the ui-host, and initializes StorageBackend), but Step 1 has a real gap:

- Step 1 says "clone the repository" but never says which one or gives the command, and the only repo linked above is logos-storage-module — which is the wrong one (that builds the module, not the UI). The UI app is in **logos-co/logos-storage-ui**, which isn't named or linked anywhere. A new reader can't actually start. Suggested the clone command inline.
- the `result/` tree in Step 1.3 doesn't match reality — it lists `bin/logos-storage-ui-app`, `logoscore`, a `modules/` dir, etc., none of which exist. The default build just produces `result/lib/storage_ui_plugin.dylib` (+ the replica factory). Same reason the "run from result/ directly: `./result/bin/logos-storage-ui-app`" line won't work — there's no result/bin; `nix run` is the way (which the doc itself says higher up).
- the "Enter development shell" code block is missing its closing fence, so it runs into the next step and mangles the rest of the page. Fixed inline.
- frontmatter: `authors` is empty and `topics` is `[]`.

I couldn't test the actual storage flow (Steps 2–5) — the connectivity setup, share, and download-by-CID all need a real display plus a router with UPnP/port-forwarding and live peers, which I don't have here. Worth Red Team dogfooding on a real network.

Calling this blocking on the Step 1 issues (clone target + the result/ tree), but they're quick fixes.

weboko (changes requested):
Dogfooded on x86_64 Linux. I could not complete Steps 3–5 (need a GUI display and live peers for the storage network), but I reviewed the doc and verified the build path.

**Unclosed code fence at Step 1.4 (still broken despite the resolved thread):**

The `nix develop` block is opened with ` ```bash ` but never closed. The thread at line 85 was resolved without the fix being applied — the code in the PR still has the unclosed fence. As a result every step after Step 1.4 (Steps 1.5, 2, 3, 4, 5, and all troubleshooting) renders inside the bash block, making the page unreadable. The fix is one line:

````suggestion
   nix develop
   ```
````

**Potentially misleading prerequisites — CMake, Ninja, pkg-config:**

The prerequisites section lists CMake (3.16+), Ninja, and pkg-config as required installs. The build path in this doc is entirely through `nix build` and `nix run`, which pull all build-time tools from the Nix closure. If these tools are actually required outside of Nix (for example, by a vendored fetch step in the flake), the doc should show a step that uses them. If they are not, listing them as prerequisites will confuse readers — especially those who just installed Nix and expect it to handle dependencies.

**Comment by kashepavadan** [comment]
@weboko could you take a manual look please? need to check UI

File: docs/storage/get-started/set-up-and-use-logos-storage-ui.md [hunk 6]
```diff
@@ -0,0 +1,185 @@
...
+- You can configure your node's network connectivity to make it reachable from the internet.
+- You can share files with other nodes and download files shared by others using a Content Identifier (CID).
+
+## Step 1: Build the application
+
+The application is built using Nix flakes. The output includes the storage UI plugin and supporting binaries.
+
+{% hint style="info" %}
+
+If Nix flakes are not enabled globally, add `experimental-features = nix-command flakes` to `~/.config/nix/nix.conf`.
```
weboko, line 44: I suggest we add following here and ensure this is replicated throughout all of the docs

Image

taken form https://github.com/logos-co/logos-docs/pull/350

docs/storage/get-started/set-up-and-use-logos-storage-ui.md
```diff
@@ -0,0 +1,185 @@
...
+1. Run the build command:
+
+   ```bash
+   nix build
+   ```
+
+   - To build individual components:
+
+   ```bash
+   # Combined plugin + QML view (default package)
```
weboko: this, I think, unnecessarily complex
let's have only command that consumer of this doc needs to use, otherwise we make their life much more complex

weboko (approved):
except comments I left, everything seemed to work!

PR #309, opened 2026-06-04
Logos ecosystem repositories reference

### LINKED ISSUE #306: add ecosystem repo map

cheny0:
Please suggest if any repos should be removed/added. Also I didn't find mixnet or peer discovery related repos so I included the anoncomms roadmap

### DIFF
--- a/docs/contribute/welcome.md
+++ b/docs/contribute/welcome.md
@@ -1,3 +1,3 @@
 # Welcome 
 
-Welcome to the Logos peer discovery documentation! We're working on the content.
+Welcome to the Contributor documentation! We're working on the content.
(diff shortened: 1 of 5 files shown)


jzaki (approved):
Handy to have all in one place.
While a repo rename will redirect, some repos were changing orgs.
Hopefully gitbook has some scripted check for broken links to catch that (other docs tools do when building).

cheny0:
> Handy to have all in one place. While a repo rename will redirect, some repos were changing orgs. Hopefully gitbook has some scripted check for broken links to catch that (other docs tools do when building).

Gitbook doesn't check external links. We plan to add CI tooling to the repo that can handle that instead

Issue #37, opened 2025-07-29
[EPIC] Create IFT Docs Standards


jorge-campo:
Docs standars provide an introduction to each type of document (concept, quickstart, procedure, etc.), informing what's the type of document about and when to use it. They're the intro to the document templates.

PR #312, opened 2026-06-04
Use the Logos Delivery module from an app

kashepavadan:
From: https://github.com/logos-co/logos-docs/pull/226


docs/messaging/delivery-module/use-the-logos-delivery-module-api-from-an-app.md
```diff
@@ -0,0 +1,209 @@
...
+owner: logos
+doc_version: 1
+slug: use-logos-delivery-module-api-from-app
+---
+
+# Use the Logos Delivery module API from an app
+
+#### Get started integrating Logos messaging into a C++ module on Testnet v0.1.
+
+This procedure covers building a Logos module that calls the Logos Delivery API to subscribe to content topics, send messages, and react to delivery events on the Logos messaging network. It gives application developers a working patter for integrating Logos messaging into their C++ modules. A complete, runnable reference implementation is available in [`logos-delivery-demo`](https://github.com/logos-co/logos-delivery-demo).
```
danisharora099:
```suggestion
This procedure covers building a Logos module that calls the Logos Delivery API to subscribe to content topics, send messages, and react to delivery events on the Logos messaging network. It gives application developers a working pattern for integrating Logos messaging into their C++ modules. A complete, runnable reference implementation is available in [`logos-delivery-demo`](https://github.com/logos-co/logos-delivery-demo).
```

docs/messaging/delivery-module/use-the-logos-delivery-module-api-from-an-app.md
```diff
@@ -0,0 +1,209 @@
...
+     ...
+   }
+   ```
+
+1. In `flake.nix`, add a matching pinned input:
+
+   ```nix
+   inputs = {
+     logos-module-builder.url = "github:logos-co/logos-module-builder";
+     delivery_module.url = "github:logos-co/logos-delivery-module/v0.1.1";
```
danisharora099: Heads up: logos-delivery-demo (the "complete reference" linked above) pins this to **v0.1.2**, not v0.1.1. Worth aligning so the doc and its reference are on the same API. Note the `send()` payload type differs between them too (see the send step below).

docs/messaging/delivery/use-logos-delivery-module-api-from-app.md
```diff
@@ -0,0 +1,209 @@
...
+   LogosResult r = m_logos->delivery_module.subscribe(contentTopic);
+   if (!r.success) {
+       qWarning() << "subscribe failed:" << r.getError();
+   }
+   ```
+
+1. Send a message. On success, `getString()` returns the request ID; track it through `messageSent` → `messagePropagated` events (or `messageError`):
+
+   ```cpp
+   LogosResult r = m_logos->delivery_module.send(contentTopic, payload);
```
danisharora099, line 169: Worth stating what `payload` is. On the pinned v0.1.1 the header is `send(QString, QString)`; the demo (v0.1.2) passes a QByteArray, and `messageReceived` hands back a QByteArray — easy to get the encoding wrong if the type isn't spelled out.

danisharora099 (commented):
Nice integration guide, and it lines up well with the actual module — I checked the documented calls and event shapes against `delivery_module_plugin.h` at the pinned commit and they match, and the `messageReceived` `data[2]` raw-bytes treatment matches what the demo actually does (good, that was a real gotcha on the earlier delivery journey).

A couple of things to fix (inline):

- the doc pins delivery_module to v0.1.1, but the logos-delivery-demo it points at as the "complete reference" is on v0.1.2. Worth aligning so people aren't reading a reference that's a version ahead.
- related: `send()`'s payload type changed between those versions (v0.1.1 header is `send(QString, QString)`; the demo passes a QByteArray). The snippet here doesn't say what `payload` is, and since receive is QByteArray it's easy to get the encoding wrong.
- typo: "a working patter" → "pattern".

One thing I couldn't get past: I tried to actually build the demo on an M-series Mac and it dies in the zerokit (RLN) dependency — cargo-vendor gets a 403 from crates.io and zerokit/liblogosdelivery-dev aren't on the logos cachix for aarch64-darwin, so it cold-builds and fails. Same missing-substitute issue I raised on the earlier delivery journey (#37). So I couldn't verify subscribe/send/receive on the live network from this doc — might be worth a prerequisites note about cachix access and pushing aarch64-darwin substitutes.

Non-blocking on the writing; the version/type mismatch is the main thing to sort.

docs/messaging/delivery/use-logos-delivery-module-api-from-app.md
igor-sirotin: Ideally we'd use `0.1.2` already, I just didn't break what I had in https://github.com/logos-co/logos-docs/pull/226.

weboko (changes requested):
Reviewed the API guide against delivery_module_plugin.h at the pinned v0.1.1 commit. Two unresolved issues from danisharora099 that are blocking:

**1. send() payload type must be stated (line 154)**

The guide documents send(contentTopic, payload) without spelling out what payload is. On v0.1.1 the signature is send(QString, QString). messageReceived hands back data[2]: QByteArray, so a reader writing code to echo received messages back will naturally try passing a QByteArray to send(), which fails to compile. The snippet needs the type explicit:

```cpp
// payload must be a QString on v0.1.1
LogosResult r = m_logos->delivery_module.send(contentTopic, payload);
```

**2. Version mismatch with linked reference (line 62)**

The flake pins delivery_module to v0.1.1, but the doc's "complete reference" (logos-delivery-demo) is on v0.1.2, which has a different send() signature. A reader who opens the demo to understand the pattern will see v0.1.2 calling send(QString, QByteArray) and the doc calling send(QString, QString) — one of them will fail. Either pin the demo to v0.1.1, or update the flake input and the send snippet to v0.1.2 (and update the type annotation accordingly).

igor-sirotin noted in the thread that they intentionally kept v0.1.1 to avoid breaking things from #226, which is fine — but then the demo link and the send type need to match that choice.

igor-sirotin:
@kashepavadan @weboko I'm happy to move on to v0.1.2, if anything. 
Just let me know where to suggest changes, I'm confused now between all the tickets and PRs 😄 

UPD: pardon, just saw https://github.com/logos-co/logos-docs/pull/226#issuecomment-4752582511. Will update the  instructions there.

xAlisher:
Verified this journey end-to-end against the real stack (ran `logos-delivery-demo`, cross-checked
`radio-basecamp`/`receiver-basecamp`). **The API reference is accurate and the documented lifecycle
works live** (createNode `logos.dev` → start → subscribe → send → propagated/sent on testnet, 8
peers). Great doc. Four fixes would let a clean-machine dev actually complete it — happy to expand any:

1. **Step 3, line 68 — wrong demo filename.** Doc says `src/delivery_module_plugin.cpp`; the demo's
   file is `src/logos_delivery_demo_plugin.cpp`. Inline suggestion below.
2. **Step 4, line 186 — `nix build .#lgx` builds an uninstallable artifact.** It emits only a
   `linux-amd64-dev` variant; `lgpm install` then fails `Package does not contain variant for
   platform: linux-x86_64` (platform side: logos-package-manager#12). The installable target is
   **`.#lgx-portable`**. Inline suggestion below.
3. **Version pin (line 19/62) vs the reference demo.** Doc pins delivery **v0.1.1**; the
   `logos-delivery-demo` flake pins **v0.1.2** (which currently fails a clean `nix build`, see
   logos-delivery-module#37). A reader copying the v0.1.1 snippet builds against a different version
   than the reference they're told to clone. Worth aligning doc and demo on one version.
4. **Troubleshooting, line 209 — `portsShift` dead-ends.** It tells you to pass `portsShift` to
   `createNode` and points to the module README for config keys, but `portsShift` isn't listed
   there. Either add it to the README config list or reference `getAvailableConfigs()`.

docs/messaging/delivery/use-logos-delivery-module-api-from-app.md
```diff
@@ -0,0 +1,203 @@
...
```
weboko, line 169: Non-blocking: the snippet never says what `payload` is. Since `messageReceived` hands `data[2]` back as a `QByteArray`, a reader echoing a received message will pass a `QByteArray` here — which is correct in v0.1.3, but worth stating so nobody reaches for a `QString`.

```suggestion
   // payload is a QByteArray of raw bytes (same type messageReceived delivers in data[2]).
   LogosResult r = m_logos->delivery_module.send(contentTopic, payload);
```

weboko (commented):
Re-reviewed after the v0.1.3 bump. Both blocking points from my earlier review are resolved:

1. **Version mismatch** — the doc now pins `delivery_module` to `v0.1.3` (794c21c) and the reference demo to tag `v0.1.0` (built against v0.1.3), so the guide and the "complete reference" are consistent.
2. **`send()` payload type** — in v0.1.3 the native signature is `send(const std::string& contentTopic, const std::vector<uint8_t>& payload)` and `messageReceived` delivers `payload` as `std::vector<uint8_t>` too (verified against [`src/delivery_module_plugin.h@794c21c`](https://github.com/logos-co/logos-delivery-module/blob/794c21cbe177bdea16d4907468eaf52d4282dda7/src/delivery_module_plugin.h)), so send and receive now use the same raw-bytes type — the earlier QString-vs-QByteArray trap is gone.

I verified every documented call (`createNode`, `start`, `subscribe`, `send`, `stop`, and the `messageReceived`/`messageSent`/`messagePropagated`/`messageError`/`connectionStateChanged` event shapes) against the pinned v0.1.3 header — all match. I did not run a full end-to-end `nix build` of the demo here (the module pulls the Nim/RLN stack from source, same cachix-substitute limitation noted on the earlier delivery journey), so this is source-verified rather than built.

One optional, non-blocking nit inline.

PR #348, opened 2026-06-17
Use Zone SDK to bridge assets

kashepavadan:
From: https://github.com/logos-blockchain/logos-blockchain/pull/2898


docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -0,0 +1,303 @@
...
+
+The Zone SDK currently supports the bundled withdrawal API only for single-sequencer Zones (`ChannelState.withdraw_threshold == 1`). Multi-sequencer Zones require building the threshold proof manually, as covered in a later step.
+
+{% endhint %}
+
+Before you start, make sure you have:
+
+- A running Logos Blockchain node reachable over HTTP
+- The Zone SDK (`lb_zone_sdk`) and `lb_core` crates added to your project
+- Completed the [Zone SDK inscription tutorial](./inscribe-data-using-the-zone-sdk.md)
```
weboko: **The prerequisite file `inscribe-data-using-the-zone-sdk.md` does not exist in the repository.**

A reader cannot complete this stated prerequisite. Either create the inscription tutorial first, or remove/replace this prerequisite until it exists.

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Step 1: Create a channel
+
+A channel is created automatically the first time an operation references a previously unseen `ChannelId`.
+
+1. Generate a [`Ed25519Key`](https://github.com/logos-blockchain/logos-blockchain/blob/master/kms/keys/src/keys/ed25519/mod.rs#L31) for your sequencer. The private key is used to sign inscriptions and channel operations, while the public key defines the `ChannelId`. An example implementation is shown below:
+
+   ```rust
+   // Generate Ed25519 key pair
+   let mut key_bytes = [0u8; ED25519_SECRET_KEY_SIZE];rand::RngCore::fill_bytes(&mut rand::thread_rng(), &mut key_bytes);
```
weboko: **Two statements are concatenated on one line — this will not compile.**

```suggestion
   let mut key_bytes = [0u8; ED25519_SECRET_KEY_SIZE];
   rand::RngCore::fill_bytes(&mut rand::thread_rng(), &mut key_bytes);
```

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -0,0 +1,303 @@
...
+   fs::write(your_file_path, key_bytes).expect("failed to write key file");
```
weboko: **`fs` is not imported — this line will not compile as written.**

`your_file_path` is also an undefined placeholder with no explanation. Suggested fix:

```suggestion
   std::fs::write(&your_key_file_path, key_bytes).expect("failed to write key file");
```

And add a note before the block explaining that `your_key_file_path` is the path where the user wants to persist the key.

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -0,0 +1,303 @@
...
+   let signing_key = Ed25519Key::from_bytes(&key_bytes)
```
weboko: **Missing semicolon — this will not compile.**

Verified against `Ed25519Key::from_bytes` signature in `kms/keys/src/keys/ed25519/mod.rs`: returns `Self` (not `Result`), so no `?` is needed, but the semicolon is required.

```suggestion
   let signing_key = Ed25519Key::from_bytes(&key_bytes);
```

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -0,0 +1,303 @@
...
+                   );
+               }
+           }
+       }
+   }
+   ```
+
+   `Event::BlocksProcessed` fires once per ingested block, whether live or backfilled, and only carries finalized items at or below LIB, so deposits surfaced here cannot be re-orged off the chain.
+
+1. Credit the user inside the Zone according to the Zone's interal state transition function rules.
```
weboko: **Typo: `interal` → `internal`**

```suggestion
1. Credit the user inside the Zone according to the Zone's internal state transition function rules.
```

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -0,0 +1,303 @@
...
+   let (result, checkpoint) = sequencer
+       .handle()
+       .submit_signed_tx(signed_tx, msg_id)?;
+   ```
+
+   Keep the result and use the returned `tx_hash` to identify the bundle. Unlike the single-sig flow, the SDK treats the caller-built transaction as opaque, so `PublishResult.tx` is `PendingTx::Inscription(InscriptionInfo)` regardless of the underlying ops.
+
+1. Match the finalized transaction by `tx_hash` once it appears in `Event::BlocksProcessed.finalized`, the same way as in the single-signature flow in Option 1.
+
+## Step 5: Recover from a reorg
```
weboko: **Step 4 is missing — the numbering jumps from Step 3 directly to Step 5.**

The document covers Step 1 (Create a channel), Step 2 (Observe deposits), Step 3 (Submit a withdrawal), then skips to Step 5 (Recover from a reorg). The reorg recovery section should be renumbered to Step 4, and the FAQ answer on line 303 that references "Step 5" should be updated to match.

```suggestion
## Step 4: Recover from a reorg
```

weboko (changes requested):
Dogfooded by reading through the tutorial as a Zone SDK user on Linux x86_64, cross-referencing every code example against the actual `zone-sdk` source in the logos-blockchain repo (crate `logos-blockchain-zone-sdk`, alias `lb-zone-sdk`). All method signatures verified. Found several blocking issues: a missing step in the numbering, a non-existent prerequisite file, two code-level bugs in the key-generation block, and a typo.

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
kashepavadan: I'm aware. This is meant to stay as a placeholder until I get that doc in

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -0,0 +1,304 @@
...
+
+1. Credit the user inside the Zone according to the Zone's internal state transition function rules.
+
+## Step 3: Submit a withdrawal
+
+The process of withdrawing funds is different for single-sequencer Zones and multiple-sequencer Zones with a withdraw threshold greater than one.
+
+### Option 1: Submit a single-sequencer withdrawal
+
+A withdrawal is initiated inside the zone and lands on-chain as a signed [`ChannelWithdraw`](https://app.notion.com/p/nomos-tech/1-5-0-Mantle-33d261aa09df8051b0d0cd4d5ddade85?source=copy_link#5de261aa09df8321b05401f2e8dea08b) operation. This step applies only when `ChannelState.withdraw_threshold == 1`.
```
weboko, line 128: Why do we keep using Notion links from our Github? Can't we have blockchain related docs on GH? 

Secondary, `nomos-tech` in the URL looks weird given we already operate using `Logos` branding. 

cc @fryorcraken

weboko (approved):

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
kashepavadan: @weboko This is where the Nomos specs currently live. I think they're being migrated at the moment.

PR #351, opened 2026-06-19
Use the Zone SDK to inscribe data on blockchain

kashepavadan:


docs/blockchain/zone-sdk/inscribe-data-on-chain-using-zone-sdk.md
```diff
@@ -0,0 +1,478 @@
...
+   pub async fn run(args: InscribeArgs) {
+
+       ...
+
+       loop {
+           tokio::select! {
+
+               // Watch for events
+               event = sequencer.next_event() => {
+                   if let Some(event) = event {
```
pradovic: next_event returns event not Option<Event>. Proposed fix: drop if let Some(event)

docs/blockchain/zone-sdk/inscribe-data-on-chain-using-zone-sdk.md
```diff
@@ -0,0 +1,478 @@
...
+           }
+       }
+
+       println!("Goodbye!");
+   }
+   ```
+
+## Step 4: Publish data
+
+Once the `ZoneSequencer` is set up, posting data to the channel is as easy as passing a vector of bytes to the sequencer handler's `publish` function. Once the transaction is posted to the network, this function returns a struct consisting of the inscription (message) ID and the current checkpoint.
```
pradovic: "Once the transaction is posted to the network" -> "Once the message is submitted to the zone-sdk, ...".

It does not mean that publish posts to the network right away, it might happen later, only local state is updated and result is returned.

pradovic (approved):
LGTM, left two comments! Thanks!

docs/blockchain/zone-sdk/inscribe-data-on-chain-using-zone-sdk.md
kashepavadan: I want to keep it consistent with the implementation in the master branch of logos-blockchain, and that's how it is there: https://github.com/logos-blockchain/logos-blockchain/blob/5069049472419e6c92cae3e1393f5d99f20b359e/deployment/tui-zone/src/lib.rs#L69

If we want to modify it, we should change it there as well, no? @pradovic

pradovic: Sorry, It was on my PR that I just merged to master (I thought I already did)

docs/blockchain/zone-sdk/inscribe-data-on-chain-using-zone-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+   ```rust
+   pub async fn run(args: InscribeArgs) {
+
+       ...
+
+       loop {
+           tokio::select! {
+
+               // Watch for events
+               event = sequencer.next_event() => {
```
weboko, line 307: **Compile error** — on the `tui-zone-tutorial` branch, `ZoneSequencer::next_event()` returns `Option<Event>`, not `Event`. Binding `event` here therefore has type `Option<Event>`, but `handle_event` on line 386 expects `Event`. This causes:

```
error[E0308]: mismatched types
   --> src/lib.rs:112:30
    |
    | handle_event(event, ...)
    |              ^^^^^ expected `Event`, found `Option<Event>`
```

Verified with `cargo check` on the tutorial branch — the error is reproducible. On `master` the signature was changed to `-> Event` directly, which is why master compiles.

Suggested fix — use the `Some(event)` pattern in `tokio::select!` to unwrap and disable the arm on `None`:

```suggestion
               Some(event) = sequencer.next_event() => {
```

docs/blockchain/zone-sdk/inscribe-data-on-chain-using-zone-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+
+   ```rust
+   // Handle sequencer events
+   //
+   // event: current sequencer event
+   fn handle_event(
+    event: Event,
+    state: &mut InMemoryZoneState,
+    sequencer: &mut ZoneSequencer<NodeHttpClient>,
+    ready_tx: &mut Option<tokio::sync::oneshot::Sender<()>>,) {
```
weboko: Non-standard formatting: the function parameters use 1-space indentation and the closing `)` is on the same line as the last argument with a leading `,`. This will pass `rustfmt` with a warning and looks inconsistent with the rest of the codebase (4-space indentation, closing `)` on its own line).

Suggested fix:

```suggestion
   fn handle_event(
       event: Event,
       state: &mut InMemoryZoneState,
       sequencer: &mut ZoneSequencer<NodeHttpClient>,
       ready_tx: &mut Option<tokio::sync::oneshot::Sender<()>>,
   ) {
```

weboko (changes requested):
Dogfooded on x86_64 Linux. Found a compile error that will block users from completing the tutorial.

The `tui-zone-tutorial` branch carries an older `zone-sdk` where `ZoneSequencer::next_event()` returns `Option<Event>` — the signature was changed to return `Event` directly on `master` after the tutorial branch was cut.

This means the tutorial code in Step 3 will not compile as written on the tutorial branch.

docs/blockchain/zone-sdk/inscribe-data-on-chain-using-zone-sdk.md
kashepavadan: Should be good now, take a look @weboko

docs/blockchain/zone-sdk/inscribe-data-on-chain-using-zone-sdk.md
```diff
@@ -0,0 +1,475 @@
...
+   git clone https://github.com/logos-blockchain/logos-blockchain.git
+   cd logos-blockchain/deployment/tui-zone
+
+   # To follow along with the tutorial
+   git checkout tui-zone-tutorial
+   ```
+
+## Step 2: Initialise the `ZoneSequencer` struct
+
+Sequencer functionality is provided via the `ZoneSequencer` struct from the Zone SDK, found in `logos-blockchain/zone-sdk/src/sequencer.rs`. When initialising this struct, provide the following arguments:
```
weboko: `logos-blockchain/zone-sdk/src/sequencer.rs` doesn't exist — `src/sequencer` is a directory, and the `ZoneSequencer` struct lives in `zone-sdk/src/sequencer/zone_sequencer.rs` (re-exported via `zone-sdk/src/sequencer/mod.rs`). Suggesting the corrected path so readers can find it:

```suggestion
Sequencer functionality is provided via the `ZoneSequencer` struct from the Zone SDK, found in `logos-blockchain/zone-sdk/src/sequencer/zone_sequencer.rs`. When initialising this struct, provide the following arguments:
```

docs/blockchain/zone-sdk/inscribe-data-on-chain-using-zone-sdk.md
```diff
@@ -0,0 +1,475 @@
...
+
+               ...
+
+           }
+       }
+
+       ...
+
+   }
+   ```
```
weboko, line 397: **Doc gap: no build/run/verify step.** I re-dogfooded this on Linux x86_64 — cloned the `tui-zone-tutorial` branch, filled in the code exactly as written in Steps 2–4, and `cargo build -p logos-blockchain-tui-zone` compiles cleanly and produces the `tui-sequencer` binary (the `[[bin]]` declared in `deployment/tui-zone/Cargo.toml`). The earlier `next_event()` mismatch is resolved: on this branch `next_event()` now returns `Event`, so the loop code here is correct. 👍

However, the tutorial ends right after the reader edits the source. "What to expect" promises they will *"publish plain-text messages as on-chain inscriptions and **verify they are finalised** in the channel"*, but there is no step that tells the reader to build/run the sequencer they just wrote, or how to verify finalisation. The reader is left with edited source and no way to run it. Suggest adding a closing step:

````suggestion
   ```

## Step 5: Build and run the sequencer

With the code filled in, build and run the sequencer from the `tui-zone` directory. The binary is named `tui-sequencer`:

    cargo run --bin tui-sequencer -- --node-url http://localhost:8080

The sequencer connects to your node, derives the channel ID from your signing key, and prints `Ready.` once it has bootstrapped. Type a message and press Enter to publish it as an on-chain inscription.

Each message first appears under **Published** (pending). Once your node finalises the block containing it, the message moves to **Finalized** — confirming the inscription is on-chain. Press Ctrl-D or enter an empty line to exit.
````

weboko (commented):
Re-dogfooded on Linux x86_64 at head `0d67087`. The completed tutorial code now **compiles cleanly** against the `tui-zone-tutorial` branch (`cargo build -p logos-blockchain-tui-zone` → `Finished`, producing the `tui-sequencer` binary), and the previously-flagged `next_event()` `Option` mismatch is resolved. One remaining gap: the tutorial never tells the reader how to build, run, or verify the sequencer they just wrote, even though "What to expect" promises verification. Inline suggestion below.

pradovic:
@kashepavadan Same question here, is this for single sequencer scenario or distributed? If it is for single, we could make it much simpler.

kashepavadan:
@pradovic This is for a single sequencer scenario. I just used the code in the logos-blockchain repo for the TUI zone - I'd like to keep both consistent, so if we simplify it here it should be simplified there as well.

pradovic:
@kashepavadan Same as in other PR, if it is single sequencer scenario it can be much simpler, just submitting. You can ignore ChannelUpdate event. Tui zone is a hybrid, it works in decentralized scenario as well. The only thing that user needs in single sequencer scenario is to keep track of published, finalized (if it needs for it's own book keeping) and save checkpoints.

Issue #343, opened 2026-06-16
[Journey] Run LEZ wallet CLI on LEZ testnet


moudyellaz:
### What the user achieves

The user can point the wallet to run on the LEZ testnet instead of the standalone sequencer

### Why it matters

A developer could test the wallet CLI also on the running LEZ testnet.

### Key components

NA

### Repository

https://github.com/logos-blockchain/logos-execution-zone/

### Runtime target

testnet v0.2

### Prerequisites

NA

### Commands and expected outputs

```sh
Inside the LEZ repo do the following:
- git checkout v0.1.2
- mv ~/.nssa/wallet ~/.nssa/wallet.old (this removes the old wallet if it already exists)
- cargo install --path wallet --force
- wallet config set sequencer_addr https://testnet.lez.logos.co
```

### Success command

wallet check-health

### Expected result

```sh
✅All looks good!
```

### Configuration details

```sh

```

### GitHub handle

@moudyellaz

### Discord handle

moudyellaz

PR #356, opened 2026-06-24
Run LEZ wallet CLI on testnet

kashepavadan:
From: https://github.com/logos-co/logos-docs/issues/343


docs/lez/get-started/connect-wallet-cli-to-lez-testnet.md
```diff
@@ -0,0 +1,64 @@
...
+- Your wallet is configured to target `https://testnet.lez.logos.co` as the sequencer address.
+
+## Install the wallet and connect it to the testnet
+
+1. In the LEZ repository, check out the correct release tag:
+
+   ```sh
+   git clone https://github.com/logos-blockchain/logos-execution-zone.git
+   cd logos-execution-zone
+   git checkout v0.1.2
```
weboko, line 31: I dogfooded this end-to-end on Linux x86_64 (cloned the repo, `git checkout v0.1.2`, `cargo install --path wallet` with the `logos-blockchain-circuits` v0.4.2 release in place, set the sequencer to the testnet, ran `wallet check-health`). **The final step fails:**

```
thread 'main' panicked at wallet/src/cli/mod.rs:134:
Local ID for authenticated transfer program is different from remote
```

The wallet connects to the testnet fine (it fetches the remote program IDs), but the `v0.1.2` build's programs don't match what the testnet runs. The intro says this targets *testnet v0.2*, yet step 1 checks out `v0.1.2`. Querying `getProgramIds` on `https://testnet.lez.logos.co` confirms a v0.2-era set (incl. a `pinata` program; `authenticated_transfer=[932356569, 2971822115, 1980043512, 429576683, ...]`) that `v0.1.2` cannot reproduce — by design `check-health` then rejects the connection.

Pin the checkout to the v0.2 release deployed on the testnet (the repo's current v0.2 line is `v0.2.0-rc1`…`v0.2.0-rc5`):

```suggestion
   git checkout v0.2.0-rc5
```

⚠️ Please confirm the **exact** RC running on `testnet.lez.logos.co` before merging — I verified `v0.1.2` fails, but the sequencer exposes no version endpoint so I couldn't auto-detect which RC it is. The wallet build must match whatever the testnet runs.

docs/lez/get-started/connect-wallet-cli-to-lez-testnet.md
```diff
@@ -0,0 +1,64 @@
...
+   ```
+
+1. Rename the existing wallet directory (if you have one) to avoid conflicts:
+
+   ```sh
+   mv ~/.nssa/wallet ~/.nssa/wallet.old
```
weboko: On a machine that has never run the wallet, `~/.nssa/wallet` doesn't exist and this aborts with `mv: cannot stat '/root/.nssa/wallet': No such file or directory` (exit 1). The step text says "if you have one", so guard the command so it's a no-op when there is nothing to move:

```suggestion
   mv ~/.nssa/wallet ~/.nssa/wallet.old 2>/dev/null || true
```

weboko (changes requested):
Dogfooded this procedure end-to-end on Linux x86_64. The build steps work, but **the documented happy path fails at `wallet check-health`** because the pinned `v0.1.2` tag is incompatible with the live v0.2 testnet. Details + suggested fixes inline. Also one minor robustness fix on the `mv` step. Nice touch that `config set sequencer_addr` auto-initializes the wallet home and that the `✅All looks good!` string matches the binary output.

docs/lez/get-started/connect-wallet-cli-to-lez-testnet.md
kashepavadan: This will work when the latest testnet version launches.

Issue #344, opened 2026-06-16
[Journey] LEZ Explorer


Arjentix:
### What the user achieves

The user has a clear representation of blockchain state including blocks, transactions and accounts.

### Why it matters

LEZ Explorer provides clear picture of what's going on on-chain. It provides view access to all committed information including account states and even account associated transactions.

### Key components

- [`LEZ Explorer`](https://github.com/logos-blockchain/logos-execution-zone/tree/main/lez/explorer_service): main component of this journey, displays information about blockchain state, connects to LEZ Indexer via RPC calls.
- [`LEZ Indexer`](https://github.com/logos-blockchain/logos-execution-zone/tree/main/lez/indexer/service): responsible for reading L2 LEZ blockchain state from L1 Logos Blockchain, provides RPC endpoints to retrieve this information.
- [`Logos Blockchain`](https://github.com/logos-blockchain/logos-blockchain): represents L1 Node and source of truth for LEZ Indexer.
- [`LEZ Sequencer`](https://github.com/logos-blockchain/logos-execution-zone/tree/main/lez/sequencer/service): accepts user transactions, executed them, combines them into block and publishes to Logos Blockchain.

### Repository

https://github.com/logos-blockchain/logos-execution-zone/

### Runtime target

testnet v0.2

### Prerequisites

- OS: Linux or macOS.
- Tools: Docker, Rust toolchain, cargo, [cargo-leptos](https://crates.io/crates/cargo-leptos), [just](https://github.com/casey/just).
- Running LEZ Indexer. By default LEZ Explorer will look for `http://localhost:8779` but user can provide it's own url using `INDEXER_RPC_URL` env var.

### Commands and expected outputs

To fully check the functionality of LEZ Explorer you'll need to have LEZ Indexer running.

And to have LEZ Indexer running you need to have Logos Blockchain Node and LEZ Sequencer running.

This may sound complicated, but we do already provide a simple way to run all of this using simple commands.

1. Clone our [LEZ repo](https://github.com/logos-blockchain/logos-execution-zone/).
2. Run `just clean` from repo derictory just to make sure there are no runtime artifacts left from previous runs you may did.
3. Then the following commands from repo directory from different terminals:
  3.1. Terminal 1: `just run-bedrock`. This will run Logos Blockchain Node (old name -- Bedrock) in Docker.
  3.2. Terminal 2: `just run-sequencer`. This will run LEZ Sequencer natively.
[...]

PR #358, opened 2026-06-24
Run lez indexer

kashepavadan:
From: https://github.com/logos-co/logos-docs/issues/349

PR #359, opened 2026-06-24
Use the LEZ Explorer

kashepavadan:
From: https://github.com/logos-co/logos-docs/issues/344


docs/lez/follow-chain/start-and-use-instance-of-lez-explorer.md
```diff
@@ -0,0 +1,81 @@
...
+{% endhint %}
+
+Before you start, make sure you have the following:
+
+- Linux or macOS operating system
+- [Docker](https://docs.docker.com/get-docker/), [Rust toolchain](https://www.rust-lang.org/tools/install), and `cargo` installed
+- [`cargo-leptos`](https://crates.io/crates/cargo-leptos) installed
+- [`just`](https://github.com/casey/just) installed
+- A local clone of the [LEZ repository](https://github.com/logos-blockchain/logos-execution-zone/)
+- A running instance of an [LEZ Indexer](./run-lez-indexer.md)
```
weboko, line 30: This relative link target doesn't exist yet: `run-lez-indexer.md` is added by #358 ("Run lez indexer"), which is still a **draft**. It's not on `main` and not in this branch, so as it stands this renders as a broken link. Worth gating this PR on #358 merging first (or landing the indexer page together), otherwise the one hard prerequisite — a running indexer — points nowhere.

weboko (commented):
Dogfooded against the upstream `logos-blockchain/logos-execution-zone` repo (`main`). Verified the commands and defaults rather than standing up the whole stack, since a live view needs bedrock (docker) + sequencer + indexer with risc0 proving running together.

**Verified correct against source:**
- `just run-explorer` exists and runs `cargo leptos serve` in `lez/explorer_service` (`Justfile`).
- Explorer serves on `http://localhost:3000` (cargo-leptos `site_addr` default; matches `explorer_service/README.md`).
- Default indexer endpoint `http://localhost:8779`, overridable via `INDEXER_RPC_URL` env or `--indexer-rpc-url` flag — exactly matches `explorer_service/src/main.rs:20` (`#[arg(long, env = "INDEXER_RPC_URL", default_value = "http://localhost:8779")]`).
- Troubleshooting port `8779` is right: the indexer's own default port is `8779` (`indexer/service/src/main.rs:13`, `default_value = "8779"`).

One blocking-ish issue inline (broken prerequisite link). Otherwise the procedure is accurate.

Note: I couldn't exercise the full "No blocks found → refresh → blocks appear" flow end-to-end here because that requires the bedrock docker node + sequencer + indexer all running (risc0 proving), which isn't feasible in this environment. The screenshots and flow are consistent with the explorer/indexer code, though.

Issue #357, opened 2026-06-24
[Journey] Start the Delivery core module using core headless — run a delivery node on Testnet v0.2


igor-sirotin:
> [!TIP]
> This is written based on https://github.com/logos-co/logos-delivery-module/blob/master/docs/run-node.md

## Outcome

A node operator has a running Logos delivery node — the `delivery_module` loaded and started on the headless `logoscore` runtime — connected to the `logos.test` network.

## Why it matters

Enables node operators to run their own Delivery node and join the network.

## Key components

- `logos-delivery-module` — the module this journey runs. Wraps `logos-delivery` and exposes module API.
- `logos-logoscore-cli` — the headless Logos daemon that hosts modules; + a CLI client that loads them and calls their methods/

## Repo

https://github.com/logos-co/logos-delivery-module @ v0.1.3

Supporting (pulled at build time by the Docker/Nix flow, not pinned by this journey): 
- https://github.com/logos-co/logos-logoscore-cli
- https://github.com/logos-co/logos-package-manager
- https://github.com/logos-messaging/logos-delivery

## Runtime target

testnet v0.2

## Prerequisites

- Linux or macOS
- Network access so the two instances can reach each other
- Tools, per chosen path: 
    - **A** Docker with Compose
    - **B** `curl` + a shell (prebuilt binaries)
    - **C** Nix with flakes enabled
    - All paths need `jq` for the verify step.

### Path A — Docker (quickest)

```bash
# Bring up the daemon container
git clone https://github.com/logos-co/logos-delivery-module.git
cd logos-delivery-module
docker compose up -d --build
# First build runs Nix + downloads packages — ~30–45 min. Later starts are fast.

# Load the module and boot the node (testnet config is mounted at /conf)
docker exec logos-node logoscore load-module delivery_module --json
docker exec logos-node logoscore call delivery_module createNode @/conf/logos-test.json --json
docker exec logos-node logoscore call delivery_module start --json
```

Verify: `docker exec logos-node logoscore status --json`. Stop: `docker compose down`.
(For the steps below, prefix CLI calls with `docker exec logos-node`.)

### Path B — Prebuilt binaries (no build, no clone)

```bash
# Install logoscore + lgpd + lgpm into ./bin
curl -fsSL https://raw.githubusercontent.com/logos-co/logos-delivery-module/master/scripts/install-node-tools.sh | sh
export PATH="$PWD/bin:$PATH"

# Fetch and install the delivery module
mkdir -p packages modules
[...]

PR #363, opened 2026-06-25
Run a delivery node on Testnet v0.2

kashepavadan:
From: https://github.com/logos-co/logos-docs/issues/357

### DIFF
--- a/docs/messaging/delivery/run-logos-delivery-node.md
+++ b/docs/messaging/delivery/run-logos-delivery-node.md
@@ -172,7 +172,7 @@ Query the node's discv5 ENR to confirm it booted with a network identity and joi
 
    ```bash
    # Path A
-   logoscore exec logos-node logoscore call delivery_module getNodeInfo MyENR --json | jq
+   docker exec logos-node logoscore call delivery_module getNodeInfo MyENR --json | jq
 
    # Paths B and C
    logoscore call delivery_module getNodeInfo MyENR --json | jq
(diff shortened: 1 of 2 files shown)


docs/messaging/delivery/run-logos-delivery-node.md
```diff
@@ -0,0 +1,212 @@
...
+
+   # Paths B and C
+   logoscore status
+   ```
+
+1. Run the [health query](https://github.com/logos-co/logos-delivery-module/blob/master/docs/query-node.md):
+
+   ```bash
+   # Path A
+   logoscore exec logos-node logoscore call delivery_module getNodeInfo MyENR --json | jq
```
weboko: `logoscore exec` is not a valid subcommand — this Path A line should use `docker exec logos-node`, like every other Path A command in the guide (Steps 2.1, 2.2, 2.3, 3.1). As written, the Path A health-query step fails for a Docker user.

```suggestion
   docker exec logos-node logoscore call delivery_module getNodeInfo MyENR --json | jq
```

weboko (commented):
Dogfooded **Path B (prebuilt binaries)** end-to-end on Ubuntu 24.04 x86_64. Worked exactly as documented: `install-node-tools.sh` fetched the three CLIs, `lgpd download delivery_module` + `lgpm install` installed `delivery_module-0.1.3`, and load-module → `createNode @logos-test.json` → `start` all returned `"success": true`. `getNodeInfo MyENR` returned a live `enr:-KW4Q…` with `"status": "ok"`, matching the expected output. `logoscore stop` shut down cleanly.

One blocking copy-paste bug on the Path A verify command (inline). Note: on Linux the prebuilt binaries are AppImages — they print a `fusermount` warning on a host without FUSE but still run via the extract-and-run fallback, so Path B's `curl` + shell prerequisites hold.

Issue #322, opened 2026-06-09
[Journey] Transfer native tokens between public and private states on LEZ inc Private Donations


moudyellaz:
### What the user achieves

A sender credits a private account (regular or PDA) they do not control, identifying the recipient only by the recipient's published (NPK, VPK) plus a chosen identifier. One reusable keypair backs up to 2^128 distinct accounts via that identifier, so a recipient can hand the same key to many senders and still receive into separate accounts. The recipient discovers and spends the funds later by scanning with `wallet account sync-private`.

### Why it matters

This is the building block for real payment flows: you can pay or donate to a private recipient without an interactive setup and without the recipient pre-registering one account per sender. Before it, one (NPK, VPK) keypair mapped to exactly one account, so a recipient could not safely receive from multiple independent senders.

### Key components

- AccountId diversification (`lee/state_machine/core/src/nullifier.rs`, `program.rs`): `for_regular_private_account(npk, identifier)` and `for_private_pda(program_id, seed, npk, identifier)` make the account id `SHA256(prefix || npk || ... || identifier)`, so one key backs many accounts; commitment/nullifier/nonce formulas now key off the account id rather than the NPK.
- Privacy-preserving circuit (`program_methods/guest/src/bin/privacy_preserving_circuit/`): the sender's account enters at visibility mask 1 (NSK known), the recipient slot at mask 2 (unauthorized); the circuit initializes the recipient account with a deterministic nullifier and encrypts the post-state to a shared secret via ephemeral ECDH against the recipient VPK.
- PrivateAccountKind header (`program.rs`): a fixed-size ciphertext header (`Regular(identifier)` vs `Pda { program_id, seed, identifier }`) the recipient decrypts to reconstruct exactly which account id an incoming transfer targets.
- Wallet account manager (`lez/wallet/src/account_manager.rs`): the `AccountIdentity` enum carries `PrivateForeign` and `PrivatePdaForeign` (know the recipient key, not the NSK) for sending, and `PrivateOwned`/`PrivatePdaOwned` for spending after discovery.
- Wallet CLI (`lez/wallet/src/cli/`): `account new private-accounts-key` publishes a reusable key, `auth-transfer send` performs the foreign transfer, `account sync-private` scans and decrypts incoming funds.

### Repository

https://github.com/logos-blockchain/logos-execution-zone

### Runtime target

testnet v0.2

### Prerequisites

[...]

PR #364, opened 2026-06-26
Transfer native tokens between public and private states on LEZ inc Private Donations

kashepavadan:
From: https://github.com/logos-co/logos-docs/issues/322

### DIFF
--- a/docs/lez/SUMMARY.md
+++ b/docs/lez/SUMMARY.md
@@ -11,6 +11,7 @@
 
 - [Transfer native tokens on the Logos Execution Zone](transfer-tokens/transfer-native-tokens-on-the-logos-execution-zone.md)
 - [Create and transfer custom tokens on the Logos Execution Zone](transfer-tokens/create-and-transfer-custom-tokens-on-the-logos-execution-zone.md)
+- [Transfer native LEZ tokens between public and private states](./transfer-tokens/transfer-native-lez-tokens-between-public-private-states.md)
 
 <!-- Keep this README at the end of the SUMMARY -->
 - [Readme](README.md)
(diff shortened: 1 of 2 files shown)


weboko (commented):
Dogfooded against the `logos-execution-zone` source at current `main`. **Every CLI command and flag in this guide is correct against the wallet source:**

- `wallet account new private-accounts-key` → `NewSubcommand::PrivateAccountsKey` (creates an npk+vpk key node, registers no account) ✓
- `wallet account show-keys --account-id <…>` → `AccountSubcommand::ShowKeys { account_id }` ✓
- `wallet auth-transfer send --from --to-npk --to-vpk --to-keys --to-identifier --amount` → `AuthTransferSubcommand::Send` with exactly those flags (`--to-keys` conflicts with `--to-npk`/`--to-vpk`, matching the doc's "instead of") ✓
- `wallet account sync-private` → `AccountSubcommand::SyncPrivate` ✓

The troubleshooting code references also check out: `check_commitments_are_new` and `check_nullifiers_are_valid` exist in `lee/state_machine/src/state.rs`, and the `AccountId = SHA256(prefix ‖ npk ‖ identifier)` claim matches `AccountId::for_regular_private_account` in `lee/state_machine/core/src/nullifier.rs` (`// 32 bytes prefix || 32 bytes npk || 16 bytes identifier`). The 2^128 identifier-space claim is consistent with `identifier: u128`.

**Could not run end-to-end:** the workspace pulls `logos-blockchain.git` as a Cargo git dependency, which this sandbox's egress policy blocks (HTTP 403), and the private-transfer path needs the Risc0 guest build plus a funded testnet wallet. So this is source-verified, not live-run. No doc changes needed from my pass.

Issue #321, opened 2026-06-09
[Journey] Allow different users to interact with same private account


moudyellaz:
### What the user achieves

Multiple users jointly control a single private account (regular or PDA) on the LEZ. From one 32-byte Group Master Secret (GMS), every member independently derives the same account keys (NSK/VSK/NPK/VPK), so any member can see and spend the shared private account without an interactive key exchange at spend time. New members are admitted by sealing the GMS to their public key and having them unseal it.

### Why it matters

It lifts private accounts from one key holder only to group-owned, which can be the foundation multisig program for k-of-n control over private state. Without it there is no way for several parties to share custody of a private balance or a private PDA.

### Key components

- GroupKeyHolder (`lee/key_protocol/src/key_management/group_key_holder.rs`): holds the GMS and deterministically derives identical per-account keys for every member; this is the cryptographic core the wallet and circuit both rely on.
- Privacy-preserving circuit (`program_methods/guest/src/bin/privacy_preserving_circuit/`): enforces authorization for private PDAs (visibility mask 3) and the tx-wide `pda_family_binding` check; consumes the keys/account-ids the GroupKeyHolder produces.
- Account id derivation (`lee/state_machine/core/src/program.rs`): `AccountId::for_private_pda(program_id, seed, npk)` binds a PDA address to the group's NPK so a PDA family belongs to one group.
- Wallet group CLI (`lez/wallet/src/cli/group.rs`, `lez/wallet/src/cli/account.rs`): user-facing commands to create groups, distribute the GMS (seal/unseal), and create shared accounts derived from a group.
- Sealing key (`SealingPublicKey`/`SealingSecretKey`, ML-KEM-768 based): a wallet's dedicated key pair for receiving a sealed GMS, kept separate from account viewing keys.

### Repository

https://github.com/logos-blockchain/logos-execution-zone

### Runtime target

testnet v0.2

### Prerequisites

- OS: Linux or macOS. On macOS the Risc0 guest build needs full Xcode with the Metal toolchain, not just command-line tools.
- Hardware: unit/integration tests run with `RISC0_DEV_MODE=1` (skips real proving) on a normal dev machine. Real proof generation needs substantial hardware.
- Tools: Rust toolchain, `cargo`, recipes in the `justfile`.
[...]

PR #366, opened 2026-06-26
Allow different users to interact with same private account

kashepavadan:
From: https://github.com/logos-co/logos-docs/issues/321

### DIFF
--- a/docs/lez/SUMMARY.md
+++ b/docs/lez/SUMMARY.md
@@ -16,5 +16,9 @@
 
 - [Start and use an instance of the LEZ Explorer](./follow-chain/start-and-use-instance-of-lez-explorer.md)
 
+## Accounts
+
+- [Set up a shared private LEZ account](./accounts/set-up-shared-private-lez-account.md)
+
 <!-- Keep this README at the end of the SUMMARY -->
 - [Readme](README.md)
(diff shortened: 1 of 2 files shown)


weboko (commented):
Dogfooded against the `logos-execution-zone` source at current `main`. **Every CLI command and flag in this guide is correct against the wallet source:**

- `wallet group new-sealing-key` → `GroupSubcommand::NewSealingKey` ✓
- `wallet group new test-group` → `GroupSubcommand::New { name }` ✓
- `wallet group list` / `ls` → `GroupSubcommand::List` (`#[command(visible_alias = "ls")]`) ✓
- `wallet group invite test-group --key <hex>` → `GroupSubcommand::Invite { name, key }` ✓
- `wallet group join my-copy --sealed <hex>` → `GroupSubcommand::Join { name, sealed }` ✓
- `wallet account new private-gms <group> [--label] [--pda --seed --program-id --identifier]` → `NewSubcommand::PrivateGms` (`--seed`/`--program-id`/`--identifier` all `requires = "pda"`, matching the doc; `--seed` doc-comment says 64-char hex = 32 bytes, so `<32-byte-hex>` is consistent) ✓

Group key derivation is real (`derive_keys_for_shared_account`, prefix `/LEE/v0.3/GroupKeyDerivation/SHA`), so the "every member derives identical NPK/VPK/NSK/VSK" claim holds, and the ML-KEM-768 sealing description matches the source.

**Could not run end-to-end:** the workspace pulls `logos-blockchain.git` as a Cargo git dependency blocked by this sandbox's egress policy (HTTP 403), and the path needs the Risc0 guest build plus a funded testnet wallet. Source-verified, not live-run. No doc changes needed from my pass.

PR #372, opened 2026-06-29
Move assets using Blockchain Dashboard

kashepavadan:
From: https://github.com/logos-co/journeys.logos.co/issues/44#issuecomment-4797971526


danisharora099 (commented):
Dogfooded this on macOS. I built and launched the app, went through the full setup (generate config → start node → it comes up and connects to peers), and cross-checked every claim in the doc against the `logos-blockchain-ui` source. The Transfer panel itself is described accurately — the flow, the fields, and the error messages all match what's in the code. A few things I'd fix:

1. **The balance doesn't update on its own after a transfer** (this is the main one). The doc says the sender balance "updates on the next balance refresh after the transfer is confirmed," which reads as automatic — but there's no periodic refresh, and a transfer doesn't trigger one. You have to hit **Refresh** in Accounts manually. Worth rewording so people aren't left staring at a stale balance.
2. **The recipient key isn't validated.** "To key (64 hex chars)" is just placeholder text — whatever you type gets passed straight to the module. The troubleshooting line ("verify the recipient key is exactly 64 hex characters") reads like the UI enforces it; it doesn't. Fine as advice, just don't imply the app checks it.
3. **Network mismatch:** the app is devnet-only (the config screen only offers Devnet / Custom config — no testnet), but the prereqs link the **testnet** faucet. Funding a testnet address against a devnet node lands you on the wrong network.
4. Minor: the "From address" dropdown lists **all** your known addresses, not only funded ones.

One heads-up that isn't really this doc's fault: I couldn't complete an actual transfer, because **the app's node can't sync devnet right now** — it starts on a different genesis than the live devnet and never syncs. I confirmed devnet itself is healthy by syncing a CLI node against it at the same time, so this is an app bug, not the doc. But it does mean this journey can't be dogfooded end to end until that's fixed — I'll raise it with the app/R&D folks separately (happy to share the details).

Not blocking the doc on the app bug — the four fixes above are what I'd want in the doc itself.

docs/blockchain/node-app/move-assets-using-logos-blockchain-app.md
```diff
@@ -0,0 +1,83 @@
...
+# Move assets using the Logos Blockchain node app
+
+#### Get started with token transfers between wallet accounts directly from the dashboard UI.
+
+This procedure covers how to send funds from one of your wallet addresses to a recipient on the Logos Blockchain using the Logos Blockchain dashboard. It is intended for users who need to move tokens between accounts without using the CLI or crafting transactions manually.
+
+Before you start, make sure you have the following:
+
+- A built [Blockchain node app](./build-and-run-logos-blockchain-node-app-ui.md)
+- At least one wallet address with a positive balance, for example one funded via the faucet at [`testnet.blockchain.logos.co`](https://testnet.blockchain.logos.co/web/faucet/)
```
weboko, line 22: **Network mismatch — still open from last round.** This links the **testnet** faucet, but the app is **devnet-only**:
- `BlockchainBackend.cpp:398` hardcodes `deployment.insert("well_known_deployment", "devnet")`
- `GenerateConfigView.qml` exposes only a **Devnet** radio (+ Custom) — no testnet option
- the app's own `README.md` funds from `https://devnet.blockchain.logos.co/web/faucet/`

Funding a testnet address against a devnet node puts the wallet on the wrong network. `https://devnet.blockchain.logos.co/web/faucet/` returns 200, so:

```suggestion
- At least one wallet address with a positive balance, for example one funded via the faucet at [`devnet.blockchain.logos.co`](https://devnet.blockchain.logos.co/web/faucet/)
```

docs/blockchain/node-app/move-assets-using-logos-blockchain-app.md
```diff
@@ -0,0 +1,83 @@
...
+   {% endhint %}
+
+## Read the transfer result
+
+1. Check the result row directly below the **Send** button:
+
+   - On success, the row displays the transaction ID or hash returned by the module. Click the copy button to copy it to the clipboard.
+   - On failure, the row displays `Error: <message>` with the rejection reason from the module.
+
+1. Confirm the sender's balance decreased by refreshing the **From address** dropdown.
```
weboko: **The From-address dropdown has no refresh — the Refresh button is in the Accounts panel.** In `TransferView.qml` the "From address" combo just renders `accountsModel`; there's no refresh affordance on it. Balances are re-fetched by the **Refresh** button in the **Accounts** view (`AccountsView.qml:48` → `refreshAccountsRequested()` → `backend.refreshAccounts()`). A transfer also doesn't auto-refresh (`transferFunds()` never calls `refreshAccounts()`; the only auto-refresh is the `QTimer::singleShot` after **Start Node**). So readers who "refresh the dropdown" will see a stale balance.

```suggestion
1. Confirm the sender's balance decreased by pressing **Refresh** in the **Accounts** panel.
```

weboko (commented):
Re-dogfooded after the "danish review" commit. I built `logos-blockchain-ui` from source per the prerequisite doc (`git clone` + `nix build` → `blockchain_ui_plugin.so`, exit 0) and cross-checked every claim in this doc against the app source. The balance-refresh wording reword (What to expect / result step) reads better now.

Two things still open that I could pin down precisely (inline):

1. **Network mismatch (your point 3 from last round).** The app is devnet-only, so the testnet faucet in the prereqs funds the wrong network.
2. **Balance refresh mechanism.** The doc points readers at the *From address dropdown* to refresh, but that control has no refresh; the **Refresh** button lives in the **Accounts** panel.

I couldn't drive an actual transfer end-to-end (headless env, no display — and the devnet-sync app bug you already logged), so these are source-verified rather than click-verified. `well_known_deployment` is hardcoded to `"devnet"` (`BlockchainBackend.cpp:398`) and `GenerateConfigView.qml` only exposes a **Devnet** radio + Custom, which corroborates your devnet-only finding. FWIW the "live balance"/"any funded wallet address" wording in *What to expect* is also slightly off (the combo lists all known addresses and the balance is only as fresh as the last manual Refresh) — minor, your call.

kashepavadan:
@weboko app should use testnet by default now. Can you check please?

docs/blockchain/node-app/move-assets-using-logos-blockchain-app.md
weboko: Update after re-dogfooding against the current app: **this is now resolved upstream — keep the `testnet` link, don't apply the devnet suggestion.**

`logos-blockchain-ui` switched from devnet to testnet in [#32 "fix(network): default"](https://github.com/logos-blockchain/logos-blockchain-ui/commit/920d9b282140ed4be4472c018ed770b11daa9566), which landed *after* my review above. That commit:
- removes the `deployment.insert("well_known_deployment", "devnet");` hardcode in `BlockchainBackend.cpp`
- removes the `devnetRadio` from `GenerateConfigView.qml`
- repoints the README faucet + block explorer to `testnet.blockchain.logos.co`

So the app is on testnet now, this doc's `testnet.blockchain.logos.co/web/faucet/` link is correct (200 OK), and the earlier testnet→devnet suggestion would reintroduce the mismatch. I'd resolve this thread.

Also re-verified the rest of the doc end-to-end against the app source on this branch's head (`4432c83`): the Transfer panel labels (`To key (64 hex chars)`, `Amount`, `Send`), the read-only balance / `---` placeholder, the result-row copy button, and the three troubleshooting strings (`Module not initialized`, `Call failed.`, `Error: …`) all match. The fresh commit's balance-refresh fix is correct — the **Refresh** button does live in the Accounts panel (`AccountsView.qml`), not on the From-address combo.

_Out of scope for this PR, but flagging for follow-up:_ the prerequisite `build-and-run-logos-blockchain-node-app-ui.md` still sends readers to the **devnet** faucet at Step 3 while the rest of that doc says testnet — worth realigning it to testnet too.

weboko:
@kashepavadan look like it, sorry, was able to check only now

PR #374, opened 2026-06-30
Bridge assets from base layer to LEZ using the blockchain dashboard

kashepavadan:
From: https://github.com/logos-co/journeys.logos.co/issues/24#issuecomment-4768386289


danisharora099 (commented):
Dogfooded on macOS and cross-checked against the app source. The wizard is documented really well — the four steps, note selection, field validation, the confirm payload, the result states, and even opcode `18` all match the app exactly. Two things I'd want fixed, plus a bigger blocker on actually running it.

1. **The tip line is cut off (line 86):** `**tip hex** or **Use query tip** (Optional) —` just ends on the dash. From the source, here's what that field actually does — suggested completion:
   > **Tip hex** or **Use query tip** (Optional) — the chain tip the deposit is built against. Leave empty to use the node's current tip, paste a specific tip hex, or click **Use query tip** to reuse the tip returned when the notes were loaded.
2. **The "Deposits are irreversible…" warning** is doc-authored, which is totally fine — just note the app doesn't show that text (the confirm step actually reads "Review the deposit. This is the exact payload that will be submitted."). Keep it as guidance, just don't imply it's on-screen.

Smaller stuff: `getChannelId` is a LEZ-node RPC, not something the app does — fine as a prerequisite, just not part of the app itself; and a couple of labels differ from the UI ("Channel ID hex" not "Channel ID (hex)", "Tip hex" not "tip hex").

The bigger issue: I couldn't run the deposit end to end, for two reasons.

- **Network story:** the app is **devnet-only** (the config screen only offers Devnet / Custom config), but this journey's channel comes from the **testnet** LEZ (`testnet.lez.logos.co`). Depositing into a testnet channel from a devnet node doesn't line up, so the network the whole flow runs on needs reconciling.
- **App can't sync:** even on devnet, the app's node currently starts on the wrong genesis and never syncs (I confirmed it's an app bug rather than devnet by syncing a CLI node against the same devnet at the same time). So the wizard can't be exercised with real funds yet.

Requesting changes for the tip line + the network story. The app-sync bug I'll raise with the app/R&D team separately.

docs/blockchain/node-app/bridge-assets-from-logos-blockchain-to-zone-using-app.md
```diff
@@ -0,0 +1,123 @@
...
+
+1. On **Step 1 – Select notes**, choose your address from **Known address…**, or paste a hex address and commit it with **Enter**.
+
+   - Notes load automatically once an address is selected. Selecting a different address reloads them.
+   
+1. Select one or more notes. **Next** enables once at least one note is selected; the full value of each selected note will be consumed.
+
+1. On **Step 2 – Fields**, enter the deposit parameters:
+
+   - **Channel ID (hex)** — the target channel.
```
weboko, line 81: The app label is `Channel ID hex` (no parens) — `ChannelDepositView.qml:292,298`:

```suggestion
   - **Channel ID hex** — the target channel.
```

docs/blockchain/node-app/bridge-assets-from-logos-blockchain-to-zone-using-app.md
```diff
@@ -0,0 +1,123 @@
...
+   - **Change public key** — prefilled from the selected wallet; receives leftover value.
+   - **Funding public keys** — one per line, prefilled; for paying the gas fee.
+   - **Max tx fee** — Maximum fee allowed to spend
+   - **Metadata (base58)** (Optional) — Usually specific to the Zone
+   - **tip hex** or **Use query tip** (Optional) — The chain tip the deposit is built against. Leave empty to use the node's current tip.
```
weboko, line 86: The app placeholder is `Tip hex` (capital T) — `ChannelDepositView.qml:407`:

```suggestion
   - **Tip hex** or **Use query tip** (Optional) — The chain tip the deposit is built against. Leave empty to use the node's current tip.
```

weboko (commented):
Re-dogfooded after the "danish review fix" commit. The tip-line completion (line 86) landed and reads well now. I verified the two runnable/buildable pieces of this journey:
- **`getChannelId` prereq works:** `curl https://testnet.lez.logos.co/ … "method":"getChannelId"` → `{"result":"0101…0101"}` (HTTP 200). 
- **App builds from source:** `git clone logos-blockchain-ui` + `nix build` → `blockchain_ui_plugin.so`, exit 0. The four wizard steps, note selection, field validation, the confirm payload, the result states, and opcode `18` all match `ChannelDepositView.qml`.

**The network story you requested changes on is still open.** I confirmed from source that the app is **devnet-only**: `BlockchainBackend.cpp:398` hardcodes `deployment.insert("well_known_deployment", "devnet")`, and `GenerateConfigView.qml` exposes only a **Devnet** radio (+ Custom) — no testnet option. So a node built from the prerequisite doc can't join the **testnet** LEZ that this journey deposits into (`testnet.lez.logos.co`), and the deposit can't be exercised end-to-end until the app gains a testnet/well-known-deployment selector (or the journey is re-scoped). Not a one-line doc fix — flagging that it's unresolved after this commit. (I couldn't drive the GUI wizard headless, so the wizard steps are source-verified rather than click-verified.)

Two small label mismatches inline (exact strings from the app, since you'd already spotted these).

kashepavadan:
@weboko app should use testnet by default now. Can you check please?

weboko:
@kashepavadan confirmed — the devnet-only blocker I flagged is resolved. I re-dogfooded against the app's latest commit (`logos-blockchain-ui` `920d9b2` — *fix(network): default* (#32)).

**What changed (and why it unblocks this journey):**
- `BlockchainBackend.cpp` no longer hardcodes `deployment.insert("well_known_deployment", "devnet")` — those 11 lines are gone. Deployment **mode 0** now omits the `deployment` block entirely and defers to the node module's own default.
- `GenerateConfigView.qml`: the old **Devnet** radio is now **Default** (pre-checked), with **Custom config** as the alternate — no more devnet-only pin.
- The same commit flips the README faucet + explorer links from `devnet.blockchain.logos.co` → `testnet.blockchain.logos.co`, confirming the intended default network is testnet.

**Dogfood results (Linux x86_64, this session):**
- ✅ Built the app from that commit end to end: `nix build` → exit 0, produced `blockchain_ui_plugin.so` (2.2 MB).
- ✅ `getChannelId` prereq against `testnet.lez.logos.co` → `0101…0101` (HTTP 200), so the LEZ channel this journey deposits into is reachable.

So a node generated with the default deployment now targets **testnet**, and the testnet LEZ deposit is no longer network-blocked.

**One remaining caveat (unchanged):** I still can't drive the Qt deposit wizard headlessly, so the four wizard steps / opcode `18` result remain *source-verified* against `ChannelDepositView.qml` rather than click-verified. Not a blocker on the doc.

Left one small inline suggestion on the launch step to point readers at the new **Default** deployment.

docs/blockchain/node-app/bridge-assets-from-logos-blockchain-to-zone-using-app.md
```diff
@@ -0,0 +1,123 @@
...
+
+- You can select one or more wallet notes and lock their full value into a channel through a guided four-step wizard.
+- You can review the exact deposit payload — channel ID, notes, keys, fee, and metadata — before submitting.
+- You can confirm the deposit succeeded by copying the returned transaction hash and finding it in a new block under opcode `18` (Channel Deposit).
+
+## Launch the app and start the node
+
+The **Operations** tab is disabled until the node is running. Notes and balances are unavailable until the node is also synced.
+
+1. Launch the app and, on the config screen, select or generate a user config and deployment config:
```
weboko: Now that the app defaults **Deployment** to **Default** = testnet (logos-blockchain-ui #32) and this journey deposits into the **testnet** LEZ (`testnet.lez.logos.co`), it's worth telling readers to keep that selection so they land on the right network — otherwise picking **Custom config** can put the node on a different deployment than the channel.

```suggestion
1. Launch the app and, on the config screen, select or generate a user config and leave **Deployment** on **Default** so the node joins **testnet** (where this journey's LEZ channel lives):
```

PR #382, opened 2026-07-06
Update wording to refer to single node with services

jzaki:
Previously phrased as different nodes, new terminology in line with node and service-discovery


cheny0 (approved):
LGTM

PR #392, opened 2026-07-10
docs: update storage documentation

2-towns:
This PR updates the documentation for Storage release 2.0.1.


kashepavadan:
@2-towns is this PR complete or will you be adding more information?

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -46,169 +46,186 @@ Before you start, make sure you have the following:
...
+    lgpm --modules-dir ./modules list
+    # storage_module appears in the listing
     ```
 
 ## Start the daemon and load the storage module
 
 Run `logoscore` with the modules directory, then load and initialise the [storage module](https://docs.logos.co/get-started/glossary#storage-module) against the testnet config.
 
-1.  Start the `logoscore` daemon in background mode:
+Several module calls in this procedure are **asynchronous**: the call returns `"result":true` as soon as the command is accepted, and the real outcome arrives later as an event in the daemon log (`storageStart`, `storageUploadDone`, `storageDownloadDone`, `storageRemoveDone`). This is why the daemon is started with its output captured to `logs.txt`, and why the steps below check that file.
```
weboko: These async completion events (`storageStart`, `storageUploadDone`, `storageDownloadDone`, `storageRemoveDone`) are real, but dogfooding this end-to-end they are **not** written to `logs.txt`. `logoscore module-info storage_module` lists them as `type: "event"` (Qt signals) delivered to event subscribers such as the Storage UI, and the `logoscore call` client never subscribes to them — grepping `logs.txt` (and the `log-file` `storage.log`) for these names returns nothing. What the steps below actually do is wait a moment and re-query with `manifests`/`exists`, so the "why the steps below check that file" rationale is misleading.

```suggestion
Several module calls in this procedure are **asynchronous**: the call returns `"result":true` as soon as the command is accepted, and the real outcome is delivered later as an event (`storageStart`, `storageUploadDone`, `storageDownloadDone`, `storageRemoveDone`). These events are emitted to event subscribers (such as the Storage UI); the `logoscore call` client does not subscribe to them, so they do **not** appear in `logs.txt`. Each step below instead waits briefly and confirms the outcome with a follow-up query (for example `manifests` or `exists`).
```

(The same "in the (daemon) log" phrasing recurs on lines 144 and 209 — separate suggestions there.)

weboko (commented):
Dogfooded this PR end-to-end on Linux, using Storage `v2.0.1` and node tools `0.2.0` (the exact versions the docs pin):

- **Build & install** (`nix build 'github:logos-co/logos-storage-module/v2.0.1#lgx-portable'` → `lgpm ... install`): works. Produces `logos-storage_module-module-lib.lgx`; `lgpm list` shows `storage_module 2.0.1`.
- **Run a Logos storage node**: full round-trip works — daemon start, `load-module`, `init`/`start`, connected to the testnet (7 bootstrap peers, all `seen:true`), `uploadUrl` → `manifests` (CID) → `downloadToUrl local=true` → `exists:true`, then `remove` → `exists:false`, `stop`/`destroy`, daemon `not_running`.
- **Run a Mix network of storage nodes**: full 6-node local network works — all nodes report identities, `mix-pool.json`/`mix-proxies.json` build correctly, node 5 upload → node 6 `downloadToUrl local=false` succeeds, and the `"Selected mix node for surbs"` grep confirms the DHT lookup was tunnelled over Mix. Content round-trips byte-for-byte.
- **Storage UI (Option B, `nix build`)**: builds; `result/lib/` contains the documented `storage_ui_plugin.so` and `storage_ui_replica_factory.so`.

Everything works end-to-end — really nice procedures. A few accuracy nits inline; the main one is that the async completion events are delivered to event subscribers (Qt signals), **not** written to `logs.txt`, which a couple of callouts imply.

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -46,169 +46,186 @@ Before you start, make sure you have the following:
...
-    </code></pre></div>
-4.  Load the storage module, initialise it with the testnet configuration, and start it:
+6.  Initialise the storage module with the testnet configuration. `init` is synchronous and returns `true` on success (the `@config.json` syntax loads the file's contents as the argument):
 
     ```sh
-    logoscore load-module storage_module
     logoscore call storage_module init @config.json
-    logoscore call storage_module start
     ```
+7.  Start the node. `start` is asynchronous: the return value only confirms the command was accepted; the real outcome is the `storageStart` event in the daemon log:
```
weboko: Same issue as line 82: the completion `storageStart` event is a Qt signal delivered to event subscribers, not a line written to the daemon log. This step actually confirms the node came up via the `debug` call in step 8.

```suggestion
7.  Start the node. `start` is asynchronous: the return value only confirms the command was accepted; completion is signalled later by the `storageStart` event (delivered to event subscribers, not written to `logs.txt`):
```

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -46,169 +46,186 @@ Before you start, make sure you have the following:
...
-    sleep 1
-
-    logoscore call storage_module exists "$(cat cid.txt)"
-    # should return false
+    # Wait a second for the removal to complete first
+    logoscore call storage_module exists "$(cat cid.txt)" | jq '.result.value'
+    # false
     ```
-3.  Stop the storage node:
+3.  Stop the storage node. `stop` is asynchronous like `start`; a `storageStop` event follows in the log. The node can be started and stopped multiple times:
```
weboko: Same as line 82: the `storageStop` event is delivered to event subscribers, not written to the log.

```suggestion
3.  Stop the storage node. `stop` is asynchronous like `start`; completion is signalled by a `storageStop` event (delivered to event subscribers, not written to `logs.txt`). The node can be started and stopped multiple times:
```

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -46,169 +46,186 @@ Before you start, make sure you have the following:
...
-    logoscore call storage_module downloadToUrl "$(cat cid.txt)" <destination-path> false <chunk-size-in-bytes>
+    logoscore call storage_module downloadToUrl "$(cat cid.txt)" "$(pwd)/logs-destination.txt" true 65536
     ```
 
-    - `downloadToUrl` takes a `local` flag which reads only from locally cached data if set to `true`.
-4. Confirm the downloaded file is present at `<output-path>` and matches the original.
+    - The `local` flag reads only from locally cached data when set to `true`; `false` fetches from the network.
 
-## Remove content and destroy the storage node
+4.  Confirm the downloaded file is present at the destination path and matches the original. You can also check the content is in local storage by CID:
```
weboko: Dogfooding gotcha: the file uploaded here is `logs.txt`, which is the **running daemon's** live stdout — it keeps growing after the upload. So a literal `diff logs.txt logs-destination.txt` reports a difference even on a perfectly good round-trip: the download matches the *snapshot* taken at upload time (`manifests` reports its `datasetSize`), which is smaller than the now-larger live log. Worth clarifying so a reader doesn't think the round-trip failed; the `exists` check below is the reliable confirmation.

```suggestion
4.  Confirm the downloaded file is present at the destination path. (`logs.txt` is the running daemon's live log and keeps growing, so it is larger than the downloaded snapshot — compare against the `datasetSize` reported by `manifests`, not the live file.) You can also check the content is in local storage by CID:
```

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -46,169 +46,186 @@ Before you start, make sure you have the following:
...
-      "listen-port": 8080,
-      "disc-port": 8090,
-      "bootstrap-node": $udp_spr_json,
-      "dht-mix-proxy": $tcp_spr_json,
-      "data-dir": "${data_dir}",
-      "mix-pool": "${mp_path}"
-    }
-    JSON
-    EOF
+    - Every omitted key keeps its default: the node joins the `logos.test` network preset (which provides the testnet bootstrap settings) and picks random ports.
```
weboko: Minor: only the TCP `listen-port` defaults to a random port (`0`). `disc-port` defaults to the fixed UDP port `8090` (per `init()`'s defaults and the [Connectivity](../concepts/connectivity.md#ports) page), not a random port — I confirmed the node bound discovery to `0.0.0.0:8090`.

```suggestion
    - Every omitted key keeps its default: the node joins the `logos.test` network preset (which provides the testnet bootstrap settings), binds discovery to the default UDP port `8090`, and picks a random TCP `listen-port`.
```

2-towns:
> @2-towns is this PR complete or will you be adding more information?

I was waiting for a review. I will apply the changes suggested and the PR should be ready.

docs/storage/get-started/run-logos-storage-node.md
2-towns: This is a good point, I think we need to change that. I will update the tutorial to create a specific file for it.

2-towns: Can you review [this commit](https://github.com/logos-co/logos-docs/pull/392/changes/899f946d1f3a3e5ad4a47ecea976f6bf226b131b) @weboko  ?

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -46,169 +46,192 @@ Before you start, make sure you have the following:
...
-4. Confirm the downloaded file is present at `<output-path>` and matches the original.
+    - The `local` flag reads only from locally cached data when set to `true`; `false` fetches from the network.
 
-## Remove content and destroy the storage node
+5.  Confirm the downloaded file is present at the destination path and matches the original. You can also check the content is in local storage by CID:
 
-To clear your local storage and destroy the storage node, follow the steps below.
+    ```sh
+    logoscore call storage_module exists "$(cat cid.txt)"
+    # returns true
```
weboko, line 198: Optional: now that the published file is a static `hello.txt`, step 5 can actually demonstrate "matches the original" with a real `diff` (this is the payoff of switching away from the live `logs.txt`). I ran exactly this and it comes back identical:

```suggestion
    # Confirm the download is a byte-for-byte copy of the original (both files are static):
    diff "$(pwd)/hello.txt" "$(pwd)/hello-destination.txt" && echo "match"
    # And confirm the content is in local storage by CID:
    logoscore call storage_module exists "$(cat cid.txt)"
    # returns true
```

weboko (approved):
**Re-reviewed the follow-up commits (incl. `899f946d`) — all my earlier findings are resolved, and I re-ran the whole procedure end-to-end. LGTM. ✅**

Dogfooded again on Ubuntu 24.04 with Storage `v2.0.1` and node tools `0.2.0`, following the doc verbatim: `nix build 'github:logos-co/logos-storage-module/v2.0.1#lgx-portable'` → `lgpm install` → daemon → `load-module` → `init`/`start`/`debug` → the **new `hello.txt` round-trip** → remove → shut down. Every step behaves as documented.

Specifically on the points from my first pass:

1. **Async events not in `logs.txt`** (steps intro, and the `start`/`stop` steps) — the reworded text is now accurate. I grepped both `logs.txt` and the `log-file` `storage.log` for `storageStart`/`storageUploadDone`/`storageDownloadDone`/`storageRemoveDone`/`storageStop`: **0 occurrences** each, matching the corrected wording.
2. **`disc-port` default `8090`** — fixed; the config-defaults bullet now reads correctly.
3. **The `logs.txt` round-trip gotcha** — `899f946d` fixes this the right way by publishing a dedicated static `hello.txt` instead of the live-growing log. I verified the round-trip: `uploadUrl` → `manifests` CID `zDvZRwzm1Ym7dNyAj3BGzPB4GbHEQRowUJbeqsKPXfrqVaN8P5Xb` → `downloadToUrl … true` → `exists: true`, and `diff hello.txt hello-destination.txt` is **byte-for-byte identical** (both 31 bytes). With the old `logs.txt` that diff would have falsely reported a mismatch, so this genuinely resolves it.

One small optional suggestion inline on the final confirm step, since the file is static now.

docs/storage/get-started/run-logos-storage-node.md
weboko: Reviewed — approved above. `899f946d` is the right call: publishing a dedicated static `hello.txt` instead of the live `logs.txt` removes the false-mismatch gotcha. I re-ran the full round-trip and `diff hello.txt hello-destination.txt` now comes back identical. 👍

kashepavadan:
Just waiting on your review now @emizzle

weboko:
human review also ✅

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -46,179 +46,204 @@ Before you start, make sure you have the following:
...
-    JSON
     EOF
-
-    chmod 755 make-mix-storage-config.sh
-    ./make-mix-storage-config.sh > config.json
     ```
-    :::
-4.  Load the storage module, initialise it with the testnet configuration, and start it:
+
+    - With `"nat": "none"` the node announces the machine's own IP as-is: it can download from the network but is not reachable from the internet from behind a router. To make it reachable, use `"nat": "upnp"` or `"nat": "extip:<your-public-IP>"` with the ports forwarded: see [Connectivity](../concepts/connectivity.md).
```
gmega, line 123: In hindsight, this is a bit nasty cause it potentially causes users to pollute the DHT with private addresses. :-) But OK after 2.1.0 this should be solved.

docs/storage/get-started/set-up-and-use-logos-storage-ui.md
```diff
@@ -28,21 +51,6 @@ Before you start, have the following ready:
 
   Verify: `nix flake --help >/dev/null 2>&1 && echo "Flakes enabled"`
 - Git with submodules support
```
gmega: Is there a git without submodules support? 😂

docs/storage/get-started/set-up-and-use-logos-storage-ui.md
gmega, line 84: Hm... do we need to talk about this here? This feels like a piece of "nix howto" embedded in our documentation.

docs/storage/get-started/set-up-and-use-logos-storage-ui.md
```diff
@@ -85,92 +86,134 @@ If Nix flakes are not enabled globally, add `experimental-features = nix-command
...
+   :::
+
+#### Build fails with HTTP 500 on BoringSSL fetch
 
-Your node must be reachable from the internet. Choose the setup option that matches your network environment.
+**Symptom:** The build fails with the following error:
+
+```
+error: Failed to fetch git repository https://boringssl.googlesource.com/boringssl : error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500
+fatal: unable to write request to remote: Broken pipe
```
gmega, line 100: Oh wow, never ran into that one 😂

docs/storage/get-started/set-up-and-use-logos-storage-ui.md
```diff
@@ -85,92 +86,134 @@ If Nix flakes are not enabled globally, add `experimental-features = nix-command
...
+```
+
+**Cause:** Git's HTTP request size limits are too low for large repositories.
+
+**Fix:** Increase the limits and retry:
+
+```bash
+git config --global http.postBuffer 524288000
```
gmega, line 108: Huh... why do you need to increase the `postBuffer` to check out a repo?

docs/storage/get-started/troubleshooting.md
```diff
@@ -0,0 +1,115 @@
...
+```sh
+sudo ufw allow <listen-port>/tcp
+sudo ufw allow 8090/udp
+```
+
+## The node was reachable, but is not anymore
+
+**Symptom**: the node worked for days or weeks, then peers suddenly cannot connect to you anymore.
+
+**Cause**: most home ISPs change your public IP from time to time. If `nat` is set to `extip:<IP>`, the node keeps announcing the old address.
```
gmega, line 101: I think the UPnP path is also broken in that sense - I don't think IP changes are handled properly. But OK, it's all going to improve soon... :-)

gmega (approved):
Very good stuff. In the end we have a UI tutorial full of screenshots (which is not the outcome of the discussion I had with @cheny0 and @fryorcraken), but we'll keep it up-to-date. :-)

docs/storage/get-started/run-logos-storage-node.md
2-towns: Yes this should be addresses in the next version

docs/storage/get-started/troubleshooting.md
2-towns: Yes I don't think so. We should be create an issue to track this.

docs/storage/get-started/set-up-and-use-logos-storage-ui.md
2-towns: Okay I am not sure about that actually.. 
Maybe it just worked for me because I retried... Should I remove it ?

kashepavadan:
@2-towns Do we still need @emizzle 's review or is it ready to merge?

2-towns:
> @2-towns Do we still need @emizzle 's review or is it ready to merge?

No I think we can merge it.

PR #398, opened 2026-07-15
fix command

vladopajic:


docs/basecamp/get-started/install-logos-basecamp.md
```diff
@@ -76,11 +76,22 @@ Internet access is required to download the binary or clone the repository, but
 1. Run the resulting binary:
 
     ```bash
-    ./result/bin/LogosBasecamp
+    ./result/logos-basecamp.AppImage
     ```
 
 ## Troubleshooting Basecamp
 
+### experimental Nix feature 'nix-command' when building appimage
```
weboko: Nit: capitalize the heading and use the product's `AppImage` casing (as elsewhere in this doc) for consistency with the other troubleshooting headings.

```suggestion
### Experimental Nix feature 'nix-command' when building the AppImage
```

docs/basecamp/get-started/install-logos-basecamp.md
```diff
@@ -76,11 +76,22 @@ Internet access is required to download the binary or clone the repository, but
...
+Following error might appear while building appimage:
```
weboko: Nit: small grammar polish.

```suggestion
The following error might appear while building the AppImage:
```

docs/basecamp/get-started/install-logos-basecamp.md
```diff
@@ -76,11 +76,22 @@ Internet access is required to download the binary or clone the repository, but
...
+
+> error: experimental Nix feature 'nix-command' is disabled; add '--extra-experimental-features nix-command' to enable it
+
+in which case command should be changed to allow flakes syntax:
```
weboko: Nit: grammar — add the article and "enable flakes" reads more naturally than "allow flakes syntax".

```suggestion
in which case the command should be changed to enable flakes:
```

weboko (commented):
**Dogfooded end-to-end — the fix is correct. ✅**

On Ubuntu 24.04 (x86_64) I installed Nix with flakes, cloned `logos-basecamp`, and followed the *Build and run from source* section:

- `nix build '.#bin-appimage'` succeeds and produces **`./result/logos-basecamp.AppImage`**.
- The path in the old text, `./result/bin/LogosBasecamp`, does **not** exist in `result/` — so this change from `./result/bin/LogosBasecamp` → `./result/logos-basecamp.AppImage` is accurate (and matches the upstream `README.md`).
- Launching the AppImage first failed with `libOpenGL.so.0: cannot open shared object file` — exactly the case the existing *libEGL.so.1 / libOpenGL.so.0 missing* entry covers. Running `apt-get install -y fuse libegl1 libopengl0` resolved it, and the app then started cleanly (`Logos Core started successfully!`).
- The new `nix-command` troubleshooting note is technically correct: the error only mentions `nix-command`, but adding `'nix-command flakes'` is right because building a flake needs both.

A few small wording/consistency nits on the new troubleshooting entry are suggested inline (non-blocking).

Issue #370, opened 2026-06-29
[Journey] Operate Payment Streams via Logos Core Module


s-tikhomirov:
## What the user achieves

The user opens a payment stream that pays a service provider continuously over time, then closes
the stream and lets the provider claim what it accrued. Concretely, the user funds a vault,
allocates part of it to a stream at a fixed rate, waits for value to accrue, closes the stream,
and the provider claims the accrued tokens. The walkthrough runs both roles on a single host with
one `logoscore` daemon and one wallet file holding two public accounts. Pause, resume, and
top-up are out of scope.

## Why it matters

Payment streams replace discrete invoices with continuous value flow, enabling real-time
compensation for ongoing services with per-second granularity. This journey demonstrates direct
operation of the LIP-155 protocol through the Logos Core module interface on a single host.

## Key components

- `payment_streams_module` — Universal Logos Core module exposing LIP-155 vault and stream lifecycle
- `logos_execution_zone` — Wallet module for chain interaction and transaction signing
- `lez-payment-streams` — On-chain SPEL guest program implementing LIP-155 semantics
- `logoscore` — Logos Core CLI for module loading and method invocation

## Repository

[https://github.com/logos-co/lez-payment-streams](https://github.com/logos-co/lez-payment-streams)

## Runtime target

Public TestNet v0.2. The walkthrough runs the LIP-155 flow on a single host against the public sequencer `https://testnet.lez.logos.co/`.

## Prerequisites

Host: Git, Nix (flakes), Docker, RISC Zero CLI (`cargo-risczero`, `rzup install`), Rust/cargo
(for `lgs setup` and for the shell to install `lgs` when it is missing), network access to the
public sequencer.

Clone the repo:

```bash
git clone https://github.com/logos-co/lez-payment-streams.git
cd lez-payment-streams
```

Initialize the shell:

```bash
chmod +x scripts/user-journey-*.sh
./scripts/user-journey-reset.sh
./scripts/user-journey-shell.sh
```

Inside the shell, run Steps 1–17. If you open a new terminal, re-run Step 1 (exports and shell helper functions) before continuing.

[...]

PR #402, opened 2026-07-17
Operate Payment Streams via Logos Core Module

kashepavadan:
From: https://github.com/logos-co/logos-docs/issues/370


docs/lez/SUMMARY.md
```diff
@@ -12,6 +12,7 @@
 - [Transfer native tokens on the Logos Execution Zone](transfer-tokens/transfer-native-tokens-on-the-logos-execution-zone.md)
 - [Create and transfer custom tokens on the Logos Execution Zone](transfer-tokens/create-and-transfer-custom-tokens-on-the-logos-execution-zone.md)
 - [Transfer native LEZ tokens between public and private states](transfer-tokens/transfer-native-lez-tokens-between-public-private-states.md)
+- [Run an LEZ payment stream](./transfer-tokens/run-lez-payment-stream.md)
```
weboko: Minor consistency nit: every other entry in this file is written without a leading `./` (e.g. `transfer-tokens/transfer-native-lez-tokens-...md`). GitBook resolves both, but for consistency with the surrounding entries:

```suggestion
- [Run an LEZ payment stream](transfer-tokens/run-lez-payment-stream.md)
```

weboko (commented):
### Dogfooding review — Run an LEZ payment stream

I walked this procedure as a new user on a clean Ubuntu 24.04 host and cross-checked every command against the source repo (`logos-co/lez-payment-streams`) and its SME-verified canonical journey (`docs/journeys/USER_JOURNEY.md`).

**What I verified working:**
- Sequencer is reachable and `getLastBlockId` returns a height (block ~29850). ✅
- `getAccount` responds and the doc's `chain_balance` parser (`sed -n 's/.*"balance":\([0-9]*\).*/\1/p'`) correctly extracts the balance from a live response. ✅
- Every referenced script exists and matches: `user-journey-reset.sh`, `user-journey-shell.sh`, `user-journey-lgs-setup.sh`, `user-journey-install-modules.sh`, `user-journey-auth-transfer.sh`. ✅
- `PAYMENT_STREAMS_GUEST_BIN` path matches the Makefile `PROGRAM_BIN`; the `program_id_hex` grep matches `fixtures/testnet-module.json`; the `make program-id` parse string matches the verified canonical journey. ✅
- All external links resolve (200): LIP-155, nixos.org/download, docs.docker.com, dev.risczero.com. ✅
- Every command block is identical to the SME-verified `USER_JOURNEY.md`, so the on-chain lifecycle is trustworthy.

**What I could not execute end-to-end here (environment limitation, not a doc defect):** the on-chain steps require building the RISC Zero guest and standing up the Nix toolchain shell, which pull the RISC Zero toolchain and the pinned `github:logos-co/*` flakes. In this sandbox those downloads are blocked by egress policy (RISC Zero release fetch fails TLS; `nix` flake tarball fetches for `logos-package-manager`/`logoscore-cli` return HTTP 403). Installed and confirmed working locally: Nix 2.34 with flakes, Docker, Rust/cargo, rsync. I'm flagging this so a reviewer on an unrestricted host runs the actual chain writes.

**Two small suggestions inline** (both non-blocking):
1. Re-add the canonical journey's note about the pinned Docker guest-builder — it's the key context for diagnosing an ImageID mismatch, which is a hard stop in this step.
2. `SUMMARY.md` link uses a `./` prefix that no other entry uses.

Nice, thorough procedure — the confirm-inclusion-before-proceeding pattern and the troubleshooting table are excellent.

docs/lez/transfer-tokens/run-lez-payment-stream.md
```diff
@@ -0,0 +1,504 @@
...
+     echo "Program id matches testnet fixture."
+     journey_ok "Guest ELF built; program id matches testnet fixture"
+   else
+     echo "Guest ImageID mismatch." >&2
+     echo "  built:    $BUILT" >&2
+     echo "  expected: $EXPECTED" >&2
+   fi
+   ```
+
+   - Stop if ImageID does not match the fixture.
```
weboko, line 160: The canonical `USER_JOURNEY.md` (Step 3) documents *why* this ImageID check is reproducible and what breaks it — the pinned Docker guest-builder image and `methods/guest/Cargo.lock`. That note was dropped here, so a reader who hits the `Guest ImageID mismatch` branch above has no guidance on the cause. Suggest restoring it:

```suggestion
   - Stop if ImageID does not match the fixture.
   - `make build` builds the guest through the pinned Docker guest-builder (`risczero/risc0-guest-builder:r0.1.88.0`) against the committed `methods/guest/Cargo.lock` and release profile — that pinning is what makes the ImageID reproducible. A mismatch almost always means one of those changed; don't run `cargo update` under `methods/guest/` unless you intend to change program identity.
```

kashepavadan:
@weboko needs a human review also

docs/lez/transfer-tokens/run-lez-payment-stream.md
```diff
@@ -0,0 +1,509 @@
...
+
+# Run an LEZ payment stream
+
+#### Get started opening and closing a continuous payment stream on testnet v0.2.
+
+This procedure covers how to open a payment stream on the LEZ that pays a service provider continuously over time and lets the provider claim the accrued tokens. Both the payer and payee roles run on a single host with one `logoscore` daemon and one wallet file holding two public accounts. Payment streams are based on the [LIP-155](https://lip.logos.co/anoncomms/raw/payment-streams.html) protocol.
+
+Before you start, make sure you have the following:
+
+- Linux host with `bash`, `git`, `python3` (stdlib only), `curl`, `rsync`
```
weboko: `rsync` is listed as a prerequisite but it isn't invoked by any command in this procedure, nor by any tool the procedure runs — I grepped the whole toolchain (`lez-payment-streams`, `scaffold`/`lgs`, `lgpm`, `logos-logoscore-cli`, `logos-execution-zone`) and found zero references to it.

Conversely, `make` **is** used in Step 1 (`make build`, `make program-id`) but isn't listed, and the Nix journey shell doesn't provide it — so a reader on a minimal host could hit `make: command not found`. Suggest swapping `rsync` for `make`:

```suggestion
- Linux host with `bash`, `git`, `python3` (stdlib only), `curl`, `make`
```

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/lez/transfer-tokens/run-lez-payment-stream.md
```diff
@@ -0,0 +1,509 @@
...
+   logoscore call payment_streams_module chainAction getStreamStatus \
+     "{\"owner\":\"$PAYER\",\"vault_id\":$VAULT_ID,\"stream_id\":$STREAM_ID}"
+   ```
+
+   - Proceed only once the read shows the stream instead of `account data missing`. If the expected state has not appeared after about one minute (several blocks), the stream creation was likely dropped — re-run the write block.
+
+1. Wait approximately 30 seconds for tokens to accrue, then check the stream status:
+
+   ```bash
+   sync_to_chain
```
weboko, line 386: Every other code block in this procedure is self-contained and copy-pasteable, but this one relies on the reader manually “waiting ~30 seconds” that only appears in the prose — pasting the block immediately after Step 5's inclusion check can read `accrued_lo` before the intended accrual window. Adding the wait to the block itself keeps it consistent with the rest of the doc and makes the `accrued_lo ≥ MIN_ACCRUED` check reliable:

```suggestion
   sleep 30
   sync_to_chain
```

---
_Generated by [Claude Code](https://claude.ai/code)_

weboko (commented):
Re-review after the three commits pushed on 2026-07-22 (`Apply suggestion from @weboko`, `merge from main`, `sasha suggestions`). The earlier note I asked for about *why* the ImageID is reproducible is now restored as the `:::info` box under Step 1 — thanks, that resolves my previous comment.

**How I dogfooded this:** I built the site (`npm run build` passes with `onBrokenLinks: 'throw'` — all links resolve, the `:::info`/`:::warning` admonitions render correctly, and the frontmatter slots in cleanly next to the other `transfer-tokens` docs). I confirmed the sequencer is reachable, and I cross-checked every `logoscore`/`payment_streams_module` command in the procedure (`create_account_public`, `account_id_to_base58`, `sync_to_block`, `initializeVault`, `deposit`, `createStream`, `getVaultStatus`, `getStreamStatus`, `closeStream`, `claim`) against the upstream `lez-payment-streams` and `logos-execution-zone` sources — all present and named correctly.

I could not execute the on-host runtime (`make build`, module install, and the chain writes) end-to-end in this sandbox: its network blocks the RISC Zero toolchain download, Nix `github:` flake fetches, and in-container git, so the guest build / `nix build` steps can't fetch their dependencies here. Those steps were validated by source inspection rather than execution.

Two small suggestions inline.

---
_Generated by [Claude Code](https://claude.ai/code)_

weboko:
@kashepavadan  human review - ✅ 

But, I personally thinks this tutorial is so complex and commands are so long that it is congenially difficult to follow. 
Maybe worth looking to improve it in next versions and I will try to think if we can upstream things into our libraries. 

For example, the fact that we have `Export session variables and define shell helper functions` for a document kind of implies something is missing.

weboko (approved):

kashepavadan:
@weboko fair enough. I'll get it published now and we can think about updating it later on

PR #406, opened 2026-07-23
doc: add guide for writing and deploying LEZ programs with logos-scaffold

### LINKED ISSUE #89: `lgs basecamp launch`: set per-profile `TMPDIR` to prevent cross-profile basecamp SIGSEGV in QRO `onClientRead`
# `lgs basecamp launch` should set per-profile `TMPDIR` to prevent cross-profile basecamp SIGSEGV

## Symptom

Running `lgs basecamp launch alice` and `lgs basecamp launch bob` concurrently on the same Linux host (so both profiles eventually load the same module — e.g. `delivery_module`) causes one or both basecamp processes to crash with `SIGSEGV` shortly after the second profile starts spawning that module's `logos_host` child. The crashing thread is always inside Qt RemoteObjects' `onClientRead` path:

```
#0  QMetaObject::methodOffset()                    ← null/dangling vtable read
#1  QMetaObject::method(int)
#2  QRemoteObjectNodePrivate::onClientRead(QObject*)
#3  doActivate<false>(QObject*, int, void**)
#6  QAbstractSocketPrivate::canReadNotification()
#7  QReadNotifier::event(QEvent*)
#8  QApplicationPrivate::notify_helper(QObject*, QEvent*)
#9  QCoreApplication::notifyInternal2(QObject*, QEvent*)
#10 socketNotifierSourceDispatch(...)
…
#16 QConnectedReplicaImplementation::waitForSource(int)
#17 LogosAPIConsumer::requestObject(...)
#18 LogosAPIConsumer::invokeRemoteMethod(...)
#19 LogosAPIClient::invokeRemoteMethod(...)
#20 LogosQmlBridge::callModule(...)               ← basecamp main_ui plugin
```

I observed this 3 times in a row across several launch orderings; the second instance to come up tends to take the first one with it within 1–2 seconds of its own modules spawning.

## Root cause (what I traced)

`liblogos_core` exposes each `logos_host` child to basecamp via a `QLocalServer`-bound socket whose name is hardcoded as `logos_token_<modulename>` (no random suffix). I confirmed this by stringsing `liblogos_core.so` (`logos_token_%1`).

Qt's `QLocalServer` resolves that name through `QStandardPaths::TempLocation`, which on Linux is the value of `TMPDIR`, falling back to `/tmp`. `lgs basecamp launch` exports `XDG_CONFIG_HOME / XDG_DATA_HOME / XDG_CACHE_HOME` per profile but does **not** export `TMPDIR`. So both profiles end up pointing at `/tmp/logos_token_<module>`.

Sequence that triggers the crash:

1. Alice launches; her `delivery_module` `logos_host` child binds `/tmp/logos_token_delivery_module`. Basecamp main_ui connects to it as a QRO replica.
2. Bob launches; his `delivery_module` `logos_host` child also tries to bind `/tmp/logos_token_delivery_module`. `QLocalServer::listen()` on the second instance unlinks the existing socket entry and binds a fresh one (Qt's `RemoveServerOnSignal`/`UnlinkOnExit` semantics).
3. Alice's `QRemoteObjectNodePrivate` still holds a replica pointing at the now-displaced socket. The next socket-read event dispatched by `QApplicationPrivate::notify_helper → QReadNotifier::event → canReadNotification` reaches `onClientRead`, which derefs a now-dangling `QMetaObject*` → SIGSEGV.

I verified the crash recurs even when bob is launched inside an unprivileged mount namespace with a private `tmpfs` over `/tmp` (`unshare -U -r -m`) — alice still crashes from her own `invokeRemoteMethod` blocking on a replica that loses its peer when bob's daemons fight over the socket name. The mount namespace isolates filesystem entries but does not save Qt from the bind-unlink race that already occurred on alice's side. Setting `TMPDIR` per profile prevents the race outright because each profile's QLocalServer binds a distinct path.

## Proposed fix

In `src/commands/basecamp.rs::launch_env`, also export `TMPDIR=<profile_dir>/xdg-tmp`, and create the directory in `seed_profiles` alongside `xdg-config / xdg-data / xdg-cache`. The directory does not need the `BASECAMP_XDG_APP_SUBPATH` ("Logos/LogosBasecampDev") suffix — Qt's `QLocalServer::listen("logos_token_<module>")` writes a flat file directly under the temp root.

Sketch:

```rust
fn launch_env(profile_dir: &Path, profile_name: &str) -> BTreeMap<String, OsString> {
    let mut env = BTreeMap::new();
    env.insert("XDG_CONFIG_HOME".into(), profile_dir.join("xdg-config").into_os_string());
    env.insert("XDG_DATA_HOME".into(),   profile_dir.join("xdg-data").into_os_string());
    env.insert("XDG_CACHE_HOME".into(),  profile_dir.join("xdg-cache").into_os_string());
    env.insert("TMPDIR".into(),          profile_dir.join("xdg-tmp").into_os_string());
    env.insert("LOGOS_PROFILE".into(),   profile_name.into());
    env
}
```

and in `seed_profiles`:

```rust
let tmp_dir = profile_dir.join("xdg-tmp");
fs::create_dir_all(&tmp_dir)
    .with_context(|| format!("create profile dir {}", tmp_dir.display()))?;
```

Test: extend `launch_env_exports_xdg_under_profile_dir_and_profile_name` to also assert `TMPDIR == /p/alice/xdg-tmp`. Cross-profile concurrent-launch should be exercisable in `DOGFOODING.md` scenario B3.

## Reproduction

- Project: `logos-battleship` (a multiplayer module that uses `delivery_module`)
- scaffold pin: a recent `feature/basecamp` branch tip
- basecamp pin: `a746cdbc521f72ee22c5a4856fd17a9802bb9d69` (v0.1.1)
- `delivery_module` pin: `1fde1566291fe062b98255003b9166b0261c6081`
- Steps:
  1. `lgs basecamp setup`
  2. Install portable `delivery_module` + `battleship` LGXs into both `alice` and `bob` profiles (same module set).
  3. In two terminals: `lgs basecamp launch alice --no-clean` and a few seconds later `lgs basecamp launch bob --no-clean`.
  4. Click *Enable Multiplayer* on either UI within ~30s of the other launching.
  5. One or both basecamp parent processes SIGSEGV. `coredumpctl info <pid>` shows the backtrace above.

## Why this is a scaffold problem and not just a basecamp bug

There is also an underlying defensiveness gap in `liblogos_core` / Qt RemoteObjects (a replica should not deref dangling memory when its peer disconnects mid-read). That belongs upstream in basecamp/Qt. But scaffold's contract — per `FURPS.md` §B3 *"Two basecamp launch invocations for different profiles on the same machine run concurrently without colliding on XDG paths, p2p identity keys, or module ports"* — already promises non-collision for the things it controls. The Qt socket directory is one more dimension that scaffold owns through env vars, and exporting `TMPDIR` per profile fulfils that promise without depending on upstream changes.

## Workaround for users on a current build

Until this lands, prefix the launch:

```sh
mkdir -p .scaffold/basecamp/profiles/alice/xdg-tmp
TMPDIR=$(realpath .scaffold/basecamp/profiles/alice/xdg-tmp) \
  lgs basecamp launch alice --no-clean
```

(repeat for bob with a separate dir).

ygd58:
Closes logos-co/scaffold#89.

Adds a step-by-step procedure guide covering:
- Project creation with `logos-scaffold new`
- Guest program structure (RISC0 zkVM, instructions, account data)
- Dev-mode build with `RISC0_DEV_MODE=1`
- Local sequencer with `localnet start`
- Deployment and program_id retrieval
- Wallet interaction and testnet deployment

Written from hands-on experience building RFP-001 (Admin Authority) and RFP-002 (Freeze Authority) PoCs using logos-scaffold.

cc @kashepavadan @fryorcraken

### DIFF
--- a/docs/blockchain/zone-sdk/write-and-deploy-a-lez-program-with-scaffold.md
+++ b/docs/blockchain/zone-sdk/write-and-deploy-a-lez-program-with-scaffold.md
@@ -1,0 +1,164 @@
+---
+title: Write and deploy a LEZ program with logos-scaffold
+doc_type: procedure
+product: blockchain
+topics:
+  - lez
+  - scaffold
+steps_layout: sectioned
+authors: ygd58
+owner: logos
+doc_version: 1
+slug: write-and-deploy-a-lez-program-with-scaffold
+sidebar_position: 4
+---
+
+# Write and deploy a LEZ program with logos-scaffold
+
+#### Use logos-scaffold to create, build, and deploy a guest program on the Logos Execution Zone testnet.
+
+[logos-scaffold](https://github.com/logos-co/scaffold) is a project scaffold and CLI tool that manages the full lifecycle of a LEZ guest program — from project creation to deployment. It pins LEZ and SPEL dependencies, builds a project-local sequencer, and handles wallet interactions, so you can focus on writing your program logic.
+
+Before you begin, ensure you have:
+
+- Linux x86\_64 or macOS
+- Rust and Cargo (latest stable)
+- Node.js v20 or later
+- A running Logos Blockchain node connected to the public testnet (see [Run a Logos Blockchain node from the CLI](../get-started/run-a-logos-blockchain-node-from-cli.md))
+
+## What to expect
+
+- You can create a new LEZ program project with logos-scaffold.
+- You can write a guest program that runs inside the RISC0 zkVM.
+- You can build and deploy your program to the LEZ testnet.
+- You can interact with your deployed program using the wallet CLI.
+
+## Step 1: Install logos-scaffold
+
+1. Download and install the latest `logos-scaffold` binary:
+
+    ```bash
+    curl -fsSL https://raw.githubusercontent.com/logos-co/scaffold/master/install.sh | sh
+    export PATH="$HOME/.logos-scaffold/bin:$PATH"
+    ```
+
+2. Verify the installation:
+
+    ```bash
+    logos-scaffold --version
+    ```
+
+## Step 2: Create a new project
+
+1. Create a new LEZ program project. Replace `my-program` with your project name:
+
+    ```bash
+    logos-scaffold new my-program
+    cd my-program
+    ```
+
+    This generates a project with the default template, which includes a sample guest program and runner scripts.
+
+2. Inspect the project layout:
+
+    ```text
+    my-program/
+    ├── scaffold.toml          # Project configuration and dependency pins
+    ├── methods/
+    │   └── guest/
+    │       └── src/bin/       # Guest programs run inside the RISC0 zkVM
+    ├── src/
+    │   └── bin/               # Runner scripts that submit transactions
+    └── .scaffold/             # Local state, wallet home, and build artifacts
+    ```
+
+## Step 3: Set up the project
+
+1. Run `setup` to sync the LEZ and SPEL repositories to their pinned commits, build the project-local sequencer and wallet binaries, and seed the default wallet:
+
+    ```bash
+    logos-scaffold setup
+    ```
+
+    This step can take several minutes on a cold cache as it builds the sequencer from source.
+
+## Step 4: Write your guest program
+
+Guest programs run inside the [RISC0 zkVM](https://dev.risczero.com/) and define the on-chain logic of your LEZ program. Each guest program in `methods/guest/src/bin/` becomes a deployable program with its own `program_id`.
+
+1. Open the sample guest program:
+
+    ```bash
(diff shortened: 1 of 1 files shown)


kashepavadan:
Replaced with https://github.com/logos-co/logos-docs/pull/410

ygd58:
Thanks for the thorough review @weboko! Added Nix to the prerequisites alongside Docker/Podman and the RISC Zero toolchain (rzup). Also updated git/curl as base requirements per your suggestion.

PR #422, opened 2026-08-03
Update LEZ wallet tutorials

kashepavadan:
Should be fine but would like an AI review


docs/lez/get-started/run-lez-wallet-via-cli.md
```diff
@@ -0,0 +1,180 @@
+---
+title: Connect the wallet CLI to the LEZ testnet
```
weboko: The frontmatter `title` still reads *Connect the wallet CLI to the LEZ testnet*, which no longer matches the page's H1 (`# Run an LEZ wallet via the CLI`) or the new slug. Docusaurus uses this field for the browser tab title and page metadata, so it should track the new title.

```suggestion
title: Run an LEZ wallet via the CLI
```

docs/lez/get-started/run-lez-wallet-via-cli.md
```diff
@@ -0,0 +1,180 @@
...
+#### Try the wallet CLI against the live LEZ testnet.
+
+This procedure explains how to install the wallet CLI from the [LEZ repository](https://github.com/logos-blockchain/logos-execution-zone/) and point it at the public [LEZ](../../get-started/glossary.md#lez) testnet sequencer.
+
+## What to expect
+
+- You can run wallet commands against the live LEZ testnet sequencer.
+- Your wallet is configured to target `https://testnet.lez.logos.co` as the sequencer address.
+
+## Install the wallet and connect it to the testnet
```
weboko, line 42: **Missing prerequisites.** Deleting `quickstart-for-the-logos-execution-zone-wallet.md` also removed its *Install the prerequisites* section, and this page now jumps straight to `cargo install --path lez/wallet --force` with nothing before it. On a clean machine that step fails: `cargo` isn't present, and the build needs a C toolchain (I had to install `clang`/`libclang-dev`/`pkg-config`/`libssl-dev` to get it to compile). Suggest reinstating a short *Before you start* section:

```suggestion
## Before you start

Make sure you have the following installed:

- **Rust and `cargo`** — install with [`rustup`](https://rustup.rs). The wallet pins its toolchain via `rust-toolchain.toml`, so the correct version is selected automatically.
- **System build dependencies** for compiling the wallet from source:

  ```bash
  # Ubuntu / Debian
  sudo apt update
  sudo apt install git curl build-essential clang libclang-dev pkg-config libssl-dev
  ```

  ```bash
  # Fedora
  sudo dnf install git curl gcc glibc-devel clang clang-devel pkgconf-pkg-config openssl-devel llvm-libs
  ```

  ```bash
  # macOS
  xcode-select --install
  brew install pkg-config openssl
  ```

## Install the wallet and connect it to the testnet
```

docs/lez/get-started/run-lez-wallet-via-cli.md
```diff
@@ -0,0 +1,180 @@
...
+   ```sh
+   git clone https://github.com/logos-blockchain/logos-execution-zone.git
+   cd logos-execution-zone
+   git checkout v0.2.0
+   ```
+
+1. Rename the existing wallet directory (if you have one) to avoid conflicts:
+
+   ```sh
+   mv ~/.nssa/wallet ~/.nssa/wallet.old 2>/dev/null || true
```
weboko: The v0.2.0 wallet stores its state under `~/.lee/wallet`, not `~/.nssa/wallet` — every command prints `Stored persistent accounts at ~/.lee/wallet/storage.json`. As written this rename targets a path that doesn't exist, so it won't actually avoid a conflict with an existing wallet. (The `~/.nssa/wallet` path referenced in the transfer tutorials is stale for the same reason.)

```suggestion
   mv ~/.lee/wallet ~/.lee/wallet.old 2>/dev/null || true
```

weboko (commented):
### Dogfooding review — LEZ wallet CLI restructure

I ran this page end to end on a clean Ubuntu 24.04 machine: cloned `logos-execution-zone` at `v0.2.0`, built and installed the wallet with `cargo install --path lez/wallet --force`, pointed it at `https://testnet.lez.logos.co`, and completed the full public flow.

**What worked (verified against the live testnet):**
- `wallet check-health` → `✅All looks good!`
- `wallet account new public` → new account, `Account is Uninitialized`
- `wallet auth-transfer init` → submitted, then `Account owned by authenticated transfer program`, `"balance":0`
- `wallet pinata claim` → `"balance":150`
- `wallet auth-transfer send --amount 37` → sender `"balance":113`, recipient `"balance":37`

So the command sequence and the expected balances (150 / 113 / 37) are all correct on testnet. 👍

The problems I hit are the prerequisites that were lost when `quickstart-for-the-logos-execution-zone-wallet.md` was deleted, plus a few sentences/paths that were copied over from that (local-sequencer) quickstart and no longer apply to a testnet-only page. Inline suggestions below.

One non-doc note: `wallet pinata claim` printed a Rust backtrace after submitting, even though the claim succeeded (balance ended at 150). That's a wallet-binary issue, not a docs one — flagging it in case it's worth an upstream report.

docs/lez/get-started/run-lez-wallet-via-cli.md
```diff
@@ -0,0 +1,180 @@
...
+In this flow, you create and initialise an account, claim testnet funds, send a transfer, and confirm resulting balances.
+
+In this task, wallet account and transfer commands interact with the authenticated-transfer [program](../../get-started/glossary.md#program), and sequencer processing determines the resulting account state. Public and private account paths share command patterns, while private paths can include local proof generation.
+
+### Create and initialise the sender public account
+
+1. Create a sender [public account](../../get-started/glossary.md#public-account) and record the `account_id` value:
+
+   ```bash
+   cd ~/logos/src/logos-execution-zone
```
weboko: `cd ~/logos/src/logos-execution-zone` points at a directory this procedure never creates — the install step above clones into the current working directory (`git clone … && cd logos-execution-zone`). Following the page from scratch, this line fails with *No such file or directory*. It's also unnecessary: `cargo install` puts `wallet` on your `PATH`, so the command runs from anywhere. (The `~/logos/src` path is a leftover from the removed quickstart, which did `mkdir -p ~/logos/src`.)

```suggestion
   wallet account new public
```

docs/lez/get-started/run-lez-wallet-via-cli.md
```diff
@@ -0,0 +1,180 @@
...
+
+   :::info
+Running `wallet auth-transfer init` initialises the sender account under the authenticated-transfer program, so the account can debit native tokens when you send transfers.
+:::
+
+   ```bash
+   wallet auth-transfer init --account-id <sender_public_account_id>
+   ```
+
+   In the output, you should see `status: "Transaction submitted"`, and the transaction hash. If you change to the terminal session where the sequencer is running, you can see a message similar to this: `Validated transaction with hash <hash_id>, including it in block`.
```
weboko: Two testnet mismatches in this paragraph:

1. The CLI prints `Transaction hash is <hash>` (plus the raw `Transaction data`), not `status: "Transaction submitted"`.
2. This page targets the public testnet sequencer, so there is no local *terminal session where the sequencer is running* — that sentence and the `Validated transaction … including it in block` log line are leftovers from the local-standalone quickstart.

```suggestion
   In the output, you should see the transaction hash printed as `Transaction hash is <hash>`.
```

docs/lez/get-started/run-lez-wallet-via-cli.md
```diff
@@ -0,0 +1,180 @@
...
+
+   ```bash
+   wallet account get --account-id <sender_public_account_id>
+   ```
+
+   In the output you should see `Account owned by authenticated transfer program`, with a `"balance":150`.
+
+### Create and fund the recipient public account
+
+1. Create a recipient public account and record the `account_id` value. Complete this step in the same terminal session as the sender account commands to avoid exporting `NSSA_WALLET_HOME_DIR` again.
```
weboko: This step tells the reader to stay in the same terminal *to avoid exporting `NSSA_WALLET_HOME_DIR` again*, but this page never exports that variable — it was a step in the removed quickstart. The clause dangles, so it's best dropped:

```suggestion
1. Create a recipient public account and record the `account_id` value:
```

PR #315, opened 2026-06-05
docs(apps): run the atomic swaps sample app locally 

danisharora099:
## Summary

This PR is rescoped from a **local source-build tutorial** to a **public-testnet journey**: a member of the public installs Logos Basecamp, adds this app's catalog URL, installs the swap app, sets up wallets in-app, takes a live offer from a headless maker we operate, and completes a real Sepolia ↔ LEZ-testnet swap ending in a receipt with explorer links.

No clone, no Nix, no local chain, no terminal.

**Why the rescope:** @kashepavadan asked on [2026-08-03](https://github.com/logos-co/logos-docs/pull/315#issuecomment-5168073648) — *"Any chance you could change this to running on testnet 0.2 rather than locally?"* That request is the reason this PR changed shape, and it is recorded here so the audit trail explains the diff.

Journey intake: #316. App-side release tracking: [logos-co/eth-lez-atomic-swaps#62](https://github.com/logos-co/eth-lez-atomic-swaps/issues/62).

## What changed

- **Audience:** developer with a toolchain → member of the public with an installed app.
- **Runtime target:** `local` → `testnet v0.2`.
- **Entry point:** `git clone --recurse-submodules` + `make setup` → add a catalog URL in Basecamp and install.
- **Wallets:** `.env` / `.env.taker` files generated by `make infra` → created and funded in-app.
- **Counterparty:** a second Basecamp window you run yourself → a headless maker we operate, discovered as a live offer.
- **Chains:** Anvil + local LEZ sequencer → public Sepolia + LEZ testnet.
- **Proof of success:** UI counters plus an optional `swap-cli` invocation → a receipt carrying explorer links.
- **Removed:** the four-terminal layout, `make swap-lgx-build`, `make basecamp-init-*`, `make basecamp-run-*`, Nix flake configuration, RISC Zero setup, and the Intel-macOS exclusion — none of which a catalog install involves.

## Where the old local flow went

The local developer flow is **not deleted** — it moves to the app's own `README.md` in `logos-co/eth-lez-atomic-swaps`.

[...]

PR #424, opened 2026-08-04
docs(basecamp): swap ETH and LEZ tokens in Logos Basecamp (public testnet rewrite, supersedes #315)

danisharora099:
## Summary

Rewrites the atomic swaps documentation packet for the **public-testnet** journey and lands it as `docs/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp.md`.

A member of the public downloads Logos Basecamp, adds a catalogue URL, installs the swap app, sets up wallets in-app, sees live offers from a maker we operate, completes a real Sepolia ↔ LEZ-testnet swap, reads a receipt with verifiable evidence, and sends feedback. No Nix, no clone, no terminal, no local chain.

**This PR supersedes #315** and should be reviewed in its place. It is opened as a separate branch rather than pushed onto `docs/atomic-swaps-sample-app-v0.1` so that a second, independent rewrite of the same packet can be compared on its merits instead of one silently overwriting the other. It branches from that branch and never force-pushes it, so #315's history is intact and preserved in this PR's commit trail.

**Why the rescope:** @kashepavadan [asked on #315](https://github.com/logos-co/logos-docs/pull/315#issuecomment-5168073648) — *"Any chance you could change this to running on testnet 0.2 rather than locally?"* The PR body of #315 was rewritten for that scope; the file never was. The published page still documented a local, build-from-source, two-Basecamp-window flow pinned to an app commit ~100 commits stale, whose entire command set (`make swap-lgx-build`, `make basecamp-init-*`, `make basecamp-run-*`) was deleted from the app months ago.

Net diff against `main` is **+290 lines across 3 files, zero deletions** — the old page was never merged to `main`, so nothing is being removed and no redirect entry is needed.

## The page

| | |
|:--|:--|
| Path | `docs/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp.md` |
| Front matter | `doc_type: procedure`, `steps_layout: sectioned`, `product: core` |
| Moved from | `docs/build-an-app/tutorials/run-the-atomic-swaps-sample-app-locally.md` |

It leaves *Build an app* for *Basecamp* because the reader now installs an app rather than building one. `sidebars.ts` drops it from Build-an-app → Tutorials and adds it to Run-an-app → Basecamp; the `basecamp` directory is autogenerated, so ordering comes from `sidebar_position: 3`.

**Outline:** Networks and addresses → How an atomic swap works → Before you begin → What to expect → Step 1 Install Logos Basecamp → Step 2 Add the catalogue and install the app → Step 3 Set up your accounts → Step 4 Check the network settings → Step 5 Take a live offer → Step 6 Read your receipt and verify it → Step 7 Send feedback → Troubleshooting → Next steps.

Written to the discipline of `run-lez-wallet-ui-and-initiate-native-token-transfers.md`: network URLs in a table up top, three paragraphs of mental model before any instruction, literal expected output at every step, one honestly-documented product wart, and a two-link Next steps.

## Ground truth, verified today

Every value in the page was checked live rather than copied forward.

| Item | Verified |
|:--|:--|
| ETH HTLC v2 `0x351B0EA07739FA9F6769213927D7836a790A5FAF` | `INTERFACE_VERSION()` → `2`, `minTimelockDelta()` → `300`, 5146 bytes of code on Sepolia |
| Old HTLC `0x8636Fe66DFee166589a913140f14d5F57394834A` | **Not dead.** Holds 4885 bytes and `minTimelockDelta()` returns 300. Only `INTERFACE_VERSION()` reverts, since v1 lacks it — that's how the app rejects it. Written out in full in the page |
| `swap` / `swap_ui` | both `0.3.3`, variants `darwin-arm64` / `linux-amd64` / `linux-arm64` |
| Display name | catalogue is **ETH ↔ LEZ Atomic Swaps**, the package is **ETH ↔ LEZ Atomic Swap** — both documented, they genuinely differ |
| LEZ explorer | `/transaction/<bare-64-hex>` returns 200, `/tx/…` returns **404** |
| LEZ sequencer | `https://testnet.lez.logos.co` reachable |

All UI labels, status strings, validation messages, and progress-stepper steps are quoted verbatim from the app and Basecamp sources rather than paraphrased.

## No "click to view on Etherscan"

`Qt.openUrlExternally` **silently no-ops inside the Basecamp host** ([eth-lez-atomic-swaps#84](https://github.com/logos-co/eth-lez-atomic-swaps/issues/84)) — a module UI cannot open a browser. Every explorer instruction on this page is therefore "click **Copy explorer link**, then paste it into your browser." This is the page's one honestly-documented wart, in Step 6.

## Gap filled

Neither `install-logos-basecamp.md` nor `install-and-load-a-module-in-logos-basecamp.md` covers adding a **third-party** catalogue URL. Step 2 documents it inline: **Package Manager → Repositories → Add a repository**, which opens **Settings → Package Repositories**. Confirmed against Basecamp `0.2.2` sources, including that **no restart is needed** for the repository itself (Basecamp re-reads the catalogue on `catalogChanged`).

## Step 3 is followable on 0.3.3 today

An earlier revision of this PR documented the guided **Setup** tab from [eth-lez-atomic-swaps#95](https://github.com/logos-co/eth-lez-atomic-swaps/pull/95) and then admitted it wasn't available, which left a reader on `0.3.3` — the only installable version — stuck halfway through. #95 is in draft and conflicting with master, so the page no longer depends on it.

Step 3 now documents the route that exists: a throwaway Ethereum key from a wallet such as MetaMask, and a LEZ account created, initialised, and funded with the LEZ wallet CLI, cross-linking the existing `run-lez-wallet-via-cli` procedure. The **Setup** tab is demoted to a clearly marked note about `0.3.4`. The rough edge is stated plainly rather than hidden, in the same register the quality-bar doc uses for its own wart, and `Before you begin` now lists Rust and `cargo` because that part genuinely needs a terminal.

**Verifying the 0.3.3 path turned up four things that would have blocked a reader**, all now fixed:

- **`Taker Account ID` was missing entirely.** Config validation requires it unconditionally, so nobody following the previous text could have reached a valid configuration at all.
- **The LEZ `Signing Key` must be 64 bare hex characters.** The UI's `isHexBytes` strips a `0x` prefix but `src/lez/client.rs` calls `hex::decode` without stripping, so a prefixed key passes validation and then fails at run time with `invalid LEZ signing key hex`. Documented as a warning.
- **`Wallet Home` is not usable by a catalogue user.** Its `.scaffold/wallet` placeholder is a repo-relative development path that doesn't exist without a clone, and the loader hard-fails with advice to "run `make setup`", which is meaningless here. The signing key is the real route.
- **Step 2 promised a seventh Setup tab** that `0.3.3` doesn't have. `AtomicSwapView.qml` at the 0.3.3 tag lists exactly six.

⚠️ **Still worth fixing app-side before #95 lands.** `SetupView.qml` and `ConfigPanel.qml` call `swapBackend.setupGenerateEthKey()`, `setupGenerateLezAccount()`, and `setupStartFunding()`, and read `setupStep` / `setupRunning` / `setupError` / `setupBalance` / `setupTarget` / `setupClaims`. But `swapBackend` is a hand-written `QtObject` façade in `swap-ui/src/qml/Main.qml` that must forward each property and slot explicitly, and #95 doesn't touch `Main.qml`. As committed, every Setup button would throw a `TypeError`. The same pre-existing gap affects `swapBackend.resetConfig()`, called by `HistoryView.qml`.

## Other discrepancies worth a reviewer's eye

- **Basecamp version.** The brief said 0.2.2; the releases page now shows **0.2.3** as latest (2026-08-04). The page points at the existing install doc rather than pinning a version, and notes the UI labels come from 0.2.2.
- **Refund copy is stale in the app.** `RefundView.qml` says "default 5m"/"default 10m" while the shipped Config defaults are 15/25 minutes. Not a docs bug, but it will confuse a reader who compares.
- **`docs/community-install.md` in the app repo** still advertises `swap v0.3.1` / `swap_ui v0.3.0` against actual `0.3.3`, and points at `docs/testnet.md` as authoritative for account setup while giving a different recipe from the one that file actually contains.
- **The LEZ wallet-home mapping** is worth an SME eye. This page routes catalogue users to the signing key precisely because `Wallet Home` looks unusable without a clone, but if the wallet CLI's own home is meant to work there, that would be the nicer path.

## What was stripped

The doc-packet artifacts that had leaked into the published page: the draft banner, the "Tested with (SME dogfood)" table, the timing table, the Red Team callout, log paths, "Known limitations", and "Clean up" — there is nothing to clean up, as the journey starts no processes and opens no terminals. GitHub `> [!NOTE]` alerts became Docusaurus `:::` admonitions.

Troubleshooting now covers only what a catalogue user can actually hit: empty Market tab, `wss://` vs `https://`, an uninitialised LEZ account, `EthLockRejected`, insufficient LEZ, the app not appearing after install, and a mid-swap timeout → the **Refund** tab, which is what makes the swap atomic. Every Nix, scaffold, `.env`, port and four-terminal row is gone.

The local developer flow is not lost — it lives in the app's own README, linked in one line under Next steps, so it can only drift in the same PR as the build system it describes.

## Gate

- `npm ci && npm run build` — **passes**, zero warnings. `onBrokenMarkdownLinks: 'throw'` is what caught the last breakage, and the relative link that broke #315's build is replaced.
- `vale` — **5 errors, down from 47.** Four are `lez` appearing inside literal URLs and one is a verbatim button label containing a spaced em dash. For comparison on `main`: the LEZ wallet quickstart has 7 and the Storage UI journey has 22.
- This PR also adds 17 domain terms to `.github/styles/config/vocabularies/Logos/accept.txt`, which incidentally fixes a pre-existing `sandboxed` error in `install-and-load-a-module-in-logos-basecamp.md`.

## Test plan

- [x] `quality:stub`
- [x] `quality:unverified` — rescoped draft with full steps, all constants verified against live chains and sources
- [ ] `quality:sme-verified` — blocked on #95 shipping as `0.3.4` so Step 3 is executable
- [ ] `quality:verified` — Red Team completes it end-to-end from a clean profile


weboko (commented):
I dogfooded this page against the **currently published** artifacts a reader would actually install — Basecamp `0.2.3`, and `swap`/`swap_ui` `0.3.3` pulled from the catalogue index — plus the live chains. Because the end-to-end GUI swap can't run headless (and the primary flow is openly blocked on the unpublished #95), I verified every constant, endpoint, and quoted UI string against ground truth rather than a live swap.

**What checks out (independently reproduced):**
- `npm ci && npm run build` passes clean; with `onBrokenMarkdownLinks: 'throw'`, that confirms every internal link and the `#basecamp`/`#catalogue` glossary anchors resolve.
- ETH HTLC `0x351B0EA07739FA9F6769213927D7836a790A5FAF` on Sepolia: `INTERFACE_VERSION()` → `2`, `minTimelockDelta()` → `300`, 5146 bytes of code. ✅
- Catalogue URL, package versions (`swap`/`swap_ui` `0.3.3`), display name **ETH ↔ LEZ Atomic Swap**, and variants `darwin-arm64`/`linux-amd64`/`linux-arm64`. ✅
- LEZ sequencer reachable; explorer `/transaction/<hash>` → 200 while `/tx/<hash>` → 404. ✅
- LEZ HTLC program `27720b5b…504070` is the value baked into the shipped `libswap_ffi.so`. ✅
- The Taker progress-stepper steps, Offer-board columns/empty-states/disclaimer, `Config ready`, `Delivery connected`, and the two copy-only receipt rows (`ETH swap ID (lock)`, `LEZ HTLC program`) all match the shipped QML verbatim. ✅

Really thorough ground-truth work. The findings below are all one theme: **the page mixes `0.3.3`-today reality with `0.3.4`-future UI, but only Step 3 is version-gated** — so Step 2's tab list and Steps 6–7's receipt/feedback controls describe a UI a reader on today's release won't find. Details inline.

docs/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp.md
```diff
@@ -0,0 +1,268 @@
...
+
+    Each opens an **Add Application** window listing **Required Packages**. Confirm with **Install** and wait for the stage label to reach `Installed`.
+
+    :::warning
+    Install `swap` before `swap_ui`. The UI package declares a dependency on the backend, and taking them the other way round leaves the interface with no backend to talk to.
+    :::
+
+1. Restart Basecamp, then open **ETH ↔ LEZ Atomic Swap** from the sidebar.
+
+    **Expected:** a row of tabs across the top: **Market**, **Config**, **Maker**, **Taker**, **Refund**, **History**, and **Setup**. Along the top you also get live `ETH` and `LEZ` balances with a **Refresh** button, and a status line that settles on `Delivery connected` once the app finds a peer.
```
weboko: **Step 2 tells the reader to expect a Setup tab that the installed version does not have — and this contradicts Step 3.**

I installed the app the way this page says: added the catalogue, took `swap` + `swap_ui` `0.3.3` (the current published versions in the index). In the shipped `swap_ui-0.3.3` package, the tab bar is a fixed model in `AtomicSwapView.qml`:

```qml
model: ["Market", "Config", "Maker", "Taker", "Refund", "History"]
```

Six tabs, **no Setup** — there is no `SetupView.qml` in the package and the string `Setup` never appears as a tab. This directly contradicts Step 3 (line 120), which correctly says *"On `0.3.3` there's no **Setup** tab."* Since Steps 1–2 install `0.3.3`, a reader following this page will not see the seventh tab this line promises.

```suggestion
    **Expected:** a row of tabs across the top: **Market**, **Config**, **Maker**, **Taker**, **Refund**, and **History**. (The **Setup** tab used in [Step 3](#step-3-set-up-your-accounts) ships only from `swap`/`swap_ui` `0.3.4`, so it isn't present on the `0.3.3` release you just installed.) Along the top you also get live `ETH` and `LEZ` balances with a **Refresh** button, and a status line that settles on `Delivery connected` once the app finds a peer.
```

docs/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp.md
```diff
@@ -0,0 +1,268 @@
...
+
+1. Click **Refresh** in the header and confirm your `LEZ` balance rose by `10`.
+
+:::info
+Offers are live announcements, not stored listings. Nothing retains them, so the **Market** tab can only show what a maker is broadcasting at that moment. The board says as much under the tape: `Offers are advertisements — a swap completes only if the maker is live.`
+:::
+
+## Step 6: Read your receipt and verify it
+
+Every finished swap writes a receipt recording both legs, so you can check the trade against the two chains instead of taking the app's word for it. Receipts persist across restarts.
```
weboko, line 227: **Steps 6–7 describe receipt controls that aren't in the published `0.3.3` — this needs the same version gate as Step 3.**

Verified against the shipped `swap_ui-0.3.3` receipt (`ReceiptCard.qml` + `TrustRow.qml`):

- There is **no `Copy explorer link` button** (line 207) and **no `Explorer link copied` confirmation** (line 209). Each reference row renders the value as a **clickable link** (`Qt.openUrlExternally(row.link)` plus a trailing `↗`), with a small **Copy** affix that copies the **raw value/hash** — not the URL. On copy, the micro-label flashes **`Copied`**.
- Because of that, line 220's *"the receipt offers **Copy value** and **Copy explorer link** buttons instead of clickable links"* is inaccurate for `0.3.3`: the rows **are** clickable links, and the only copy control is labelled **Copy** and yields the bare hash, so "paste the link into your browser" would paste `0x…`, not a `.../tx/0x…` URL.
- **`Copy feedback link`** and **`Copy safe evidence`** (Step 7, line 229) don't exist in `0.3.3` either — no `feedback`/`evidence` button strings are in the package.

This looks like the same unpublished work as Step 3 (the in-app Setup/receipt overhaul), but only Step 3 carries the `0.3.4` admonition. Suggest gating Steps 6–7 the same way (or aligning the labels to the `0.3.3` receipt):

```suggestion
:::warning
Like [Step 3](#step-3-set-up-your-accounts), this step and [Step 7](#step-7-send-feedback) describe receipt controls — **Copy explorer link**, **Copy value**, the `Explorer link copied` confirmation, **Copy feedback link**, and **Copy safe evidence** — that aren't in the published `0.3.3`. On `0.3.3` the receipt shows each reference as a clickable link with a small **Copy** affix that copies the raw value (the label flashes `Copied`). Gate these two steps to the release that introduces the dedicated buttons, or align the labels with the shipped `0.3.3` receipt.
:::

Every finished swap writes a receipt recording both legs, so you can check the trade against the two chains instead of taking the app's word for it. Receipts persist across restarts.
```

docs/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp.md
```diff
@@ -0,0 +1,268 @@
...
+
+The **Config** tab holds every endpoint, address, and key the app uses, grouped under **Ethereum**, **LEZ**, and **Swap parameters**. Almost all of it is already correct. This step is a check rather than a data-entry exercise.
+
+1. Open the **Config** tab.
+
+1. Under **Ethereum**, confirm **RPC URL** is `wss://ethereum-sepolia-rpc.publicnode.com` and **HTLC Contract Address** is `0x351B0EA07739FA9F6769213927D7836a790A5FAF`. **Private Key** and **Recipient Address** were filled in by [Step 3](#step-3-set-up-your-accounts).
+
+1. Under **LEZ**, confirm **Sequencer URL** is `https://testnet.lez.logos.co` and **HTLC Program ID** is `27720b5b0345135d8e684eb172c27f5fb237548cc891a3ec889d0ed340504070`.
+
+1. Leave **Swap parameters** alone.
```
weboko, line 185: **Minor: verbatim label casing.** The shipped `ConfigPanel.qml` labels this section **`Swap Parameters`** (capital P); the page writes **`Swap parameters`** here and on line 153. Given the page's stated discipline of quoting UI labels verbatim, worth matching the capital P in both places.

```suggestion
1. Leave **Swap Parameters** alone.
```

kashepavadan:
Let's wait until https://github.com/logos-co/eth-lez-atomic-swaps/pull/95 is merged in. What do you think @danisharora099?

Issue #403, opened 2026-07-20
[Journey] Register an RLN membership on the LEZ testnet


adklempner:
# Journey: Register an RLN membership on LEZ testnet

### What the user achieves

Installs the RLN Membership app in Basecamp and uses it to register a
membership on LEZ testnet.

### Why it matters

Demonstrates the ability to acquire RLN memberships for participating in
DoS-protected networks on the Logos stack

### Key components

- **Logos Basecamp** — the desktop host with the package manager the app is
  installed from.
- **RLN Membership (`rln_membership_ui`)** — the onboarding wizard and
  membership status card.
- **RLN modules (`liblogos_rln_module`, `liblogos_lez_rln_module`) and wallet
  (`lez_core`)** — the backend; installed automatically as dependencies.

### Repository

[https://github.com/logos-co/logos-rln-modules](https://github.com/logos-co/logos-rln-modules) (`main`)
— reference only; nothing needs to be cloned or built for this flow.

### Runtime target

testnet v0.2.2

### Prerequisites

  - **macOS (Apple Silicon)** — `.dmg`, or **Linux (x86-64 / aarch64)** —
    `.AppImage`.
  - **Network access** to the testnet sequencer `https://testnet.lez.logos.co/`.
  - **Disk**: ~1 GB free.

### Commands and expected outputs

```sh
> The wizard runs ~5–20 minutes end to end; syncing dominates and varies with
> testnet throughput.

**a. Install Basecamp** from the latest release:

https://github.com/logos-co/logos-basecamp/releases/latest

macOS: open the `-aarch64.dmg` and drag Logos Basecamp to Applications. If
Gatekeeper blocks the first launch, right-click the app → Open.
Linux: download the `.AppImage` for your arch, `chmod +x` it, and run it.

Expected: the Basecamp window opens with a sidebar (Applications, Package
Manager, Settings).

**b. Add the RLN module registry.** Open **Package Manager** →
**Add repository** and paste:

https://github.com/logos-co/logos-rln-modules/releases/download/index/logos-repo.json

Expected: the repository lists as **Logos RLN Membership** with eight packages.

**c. Install the app.** Find **RLN Membership** and click **Install** →
**Install with dependencies**.

Expected: RLN Membership appears in the sidebar.

**d. Register a membership.** Open **RLN Membership** from the sidebar:
1. Click **Get started** on the welcome screen
2. Watch the single progress bar run **Syncing → Claiming → Registering**

[...]

PR #427, opened 2026-08-05
Register an RLN membership on the LEZ testnet

kashepavadan:
From: https://github.com/logos-co/logos-docs/issues/403


docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,90 @@
...
+
+## What to expect
+
+- You can install Basecamp from a release binary and add the RLN module registry without building anything from source.
+- You can register a new RLN membership through a guided wizard that syncs your wallet, claims faucet funds, and confirms registration automatically.
+- You can verify your membership's state, leaf index, and registration transaction, and look it up in the block explorer.
+
+## Step 1: Install the RLN Membership app
+
+Add the RLN module registry, then install the app with its dependencies from both the registry and the official catalog.
```
weboko, line 33: House style (Vale `Logos.BritishSpelling`, error level) flags **catalog** → **catalogue** here, and again on line 45.
```suggestion
Add the RLN module registry, then install the app with its dependencies from both the registry and the official catalogue.
```

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,90 @@
...
+
+1. In Basecamp's **Package Manager**, click **Add repository** and paste the registry URL:
```
weboko, line 35: On Basecamp v0.2.3 there is no **Add repository** button in Package Manager. The control is **Manage Repositories** (top-right), which opens **Settings › Package Repositories**; you paste the URL under **Add a repository** and click **Add**. After adding, click **Reload** back in Package Manager — the new packages don’t appear in the list until you do.
```suggestion
1. In Basecamp's **Package Manager**, click **Manage Repositories**, then under **Add a repository** paste the registry URL and click **Add**:
```

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,90 @@
...
+
+   ```text
+   https://github.com/logos-co/logos-lez-rln/releases/download/index/logos-repo.json
+   ```
+
+   - The repository lists as **Logos RLN Membership** with seven packages.
+
+1. Find **RLN Membership** in the list and click **Install**, then click **Install with dependencies**.
```
weboko, line 43: ⛔ **Blocker — the documented install does not complete on Basecamp v0.2.3.** Two issues:

**(a) There is no “Install with dependencies” action.** Clicking **Install** opens an **Install Package?** dialog with a single **Install** button that reads *“Install 'rln_membership_ui' v0.7.0. No other packages need to change.”* It installs only `rln_membership_ui` and does **not** pull the dependencies.

**(b) The app then can’t launch.** With the backend modules absent, opening **RLN Membership** fails with:
```
[warning] [logos] Module not found in known modules: logos_execution_zone
[warning] [logos] Missing dependencies detected: logos_execution_zone
[warning] [logos] Cannot resolve dependencies for: logos_execution_zone
```
**Root cause:** `rln_membership_ui` v0.7.0 (and `liblogos_rln_module` v1.1.0) list a dependency on **`logos_execution_zone`**, but no package by that name exists in the official catalogue (which provides **`lez_core`** — *“LEZ Core Module / Logos Execution Zone Core Module”*) or in the RLN registry, so the dependency is unresolvable.

This lives in the upstream `logos-co/logos-lez-rln` registry, so it isn’t a change to this file — but it blocks the happy path. Suggested resolution: regenerate the RLN manifests to depend on `lez_core` (or publish/alias the module as `logos_execution_zone`), then align this step with the actual UX. Once the names match, please confirm whether Basecamp auto-installs the deps or whether the user must first install `lez_core`, `libp2p_module`, `liblogos_rln_module`, `liblogos_rln_membership_module`, and `keycard_capture_module` — the step should say which.

At minimum the non-existent button name should go (this only fixes the label; the dependency blocker above still needs the upstream fix):
```suggestion
1. Find **RLN Membership** in the list and click **Install**, then confirm in the **Install Package?** dialog.
```

weboko (commented):
I dogfooded this end-to-end on Linux (x86_64) with **Logos Basecamp v0.2.3** (release AppImage). Install and launch worked once I installed `libegl1 libopengl0` (the `libOpenGL.so.0` error the install guide's troubleshooting covers). Adding the registry and reloading surfaced the **Logos RLN Membership** repo with all **7 packages**, exactly as documented, and the external endpoints in the prerequisites all resolve (sequencer, explorer, registry index).

I hit one **hard blocker** at Step 1 and a couple of **UI-label mismatches** against Basecamp 0.2.3. Details inline with suggested changes.

**Summary of the blocker:** the **RLN Membership** app installs, but it **cannot launch** — opening it logs `Cannot resolve dependencies for: logos_execution_zone`. `rln_membership_ui` (and `liblogos_rln_module`) declare a dependency on `logos_execution_zone`, which does not exist by that name in either the official catalogue (it ships as `lez_core` — “LEZ Core Module”) or the RLN registry. This is an upstream `logos-lez-rln` manifest issue rather than a doc-text fix, but it stops the documented procedure from completing end-to-end, so flagging it here before this leaves draft.

kashepavadan:
@weboko needs an updated review

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,89 @@
...
+
+## What to expect
+
+- You can install Basecamp from a release binary and add the RLN module registry without building anything from source.
+- You can register a new RLN membership through a guided wizard that syncs your wallet, claims faucet funds, and confirms registration automatically.
+- You can verify your membership's state, leaf index, and registration transaction, and look it up in the block explorer.
+
+## Step 1: Install the RLN Membership app
+
+Add the RLN module registry, then install the app with its dependencies from both the registry and the official catalog.
```
weboko: `catalog` trips `Logos.BritishSpelling`, which is **level: error** in `.github/styles/Logos/BritishSpelling.yml` (`catalog: catalogue`, `ignorecase: true`), so this fails the Vale CI job.

```suggestion
Add the RLN module registry, then install the app with its dependencies from both the registry and the official catalogue.
```

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. In Basecamp's **Package Manager**, click **Add repository** and paste the registry URL:
+
+   ```text
+   https://github.com/logos-co/logos-rln-modules/releases/download/index/logos-repo.json
+   ```
+
+   - The repository lists as **Logos RLN Membership** with eight packages.
```
weboko: ⛔ **There is no `Add repository` button in Package Manager on Basecamp v0.2.3.** I raised this last review and it's unchanged, so a reader still stalls on the first instruction of Step 1.

The three controls at the top-right of Package Manager are **Install Local Package**, **Reload**, and **Manage Repositories**. The real path is: **Manage Repositories** → lands on **Settings › Package Repositories** → paste under **Add a repository** → click **Add**.

The **Reload** step also has to be called out: after adding the repo I went back to Package Manager and the 8 new packages were *not* in the list until I clicked **Reload**.

````suggestion
1. In Basecamp's **Package Manager**, click **Manage Repositories** (top-right) to open **Settings › Package Repositories**.

1. Under **Add a repository**, paste the registry URL and click **Add**:

   ```text
   https://github.com/logos-co/logos-rln-modules/releases/download/index/logos-repo.json
   ```

   - The repository is added as **Logos RLN Membership** with eight packages.

1. Return to **Package Manager** and click **Reload**.

   - The new packages do not appear in the list until you reload.
````

File: docs/lez/rln/register-rln-membership-from-basecamp.md [hunk 6]
```diff
@@ -0,0 +1,89 @@
...
+
+1. Find **RLN Membership** in the list and click **Install**, then click **Install with dependencies**.
+
+   - **RLN Membership** appears in the sidebar once the install completes.
```
weboko: ⛔ **There is no `Install with dependencies` button**, and one Install click is not enough to get the app installed. Both halves of this line are wrong on v0.2.3.

What actually happens:

**(a) The dialog has a single `Install` button.** Clicking **Install** opens **Install Package?** reading *"Install 'rln_membership_ui' v0.7.2. Installing it also applies the dependency changes listed below"*, listing all six deps (`keycard_capture_module`, `rln_gifter_module`, `libp2p_module`, `liblogos_lez_rln_module`, `liblogos_rln_module`, `lez_core`). The only buttons are **Cancel** and **Install**. Good news: unlike last review, dependency resolution is now correct.

**(b) The first Install installs only the dependencies — not the app.** After it finished, the six deps were all `INSTALLED` but **RLN Membership** still showed `INSTALL`, and only `modules/` was populated. I had to click **Install** a second time (the dialog then said *"No other packages need to change"*) before `rln_membership_ui` was fetched into `plugins/`.

**(c) It does not appear in the sidebar until Basecamp is restarted.** After the second install the row went to a red `RETRY`, and no sidebar entry appeared. The files were complete on disk, and restarting Basecamp loaded all six modules and put **RLN Membership** in the sidebar. The `RETRY` badge is misleading — the install had in fact succeeded.

```suggestion
1. Find **RLN Membership** in the list and click **Install**. In the **Install Package?** dialog, review the listed dependency changes and click **Install** to confirm.

   - The wallet, libp2p, and RLN modules are installed for you — `libp2p_module` resolves from the official catalogue and the rest from the registry you added.
   - This first pass installs the dependencies only. When it finishes, click **Install** on **RLN Membership** again to install the app itself; the dialog confirms **No other packages need to change**.

1. Restart Basecamp.

   - **RLN Membership** appears in the sidebar after the restart. If the row shows **RETRY** before you restart, the install has still succeeded — restart and check the sidebar before retrying.
```

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Step 2: Register a membership
+
+Open the RLN Membership app and complete the guided wizard to register a new membership.
+
+:::info
+The **Registering** screen runs two unattended waits: a first-run chain sync (about 5–6 minutes for a fresh wallet, roughly 50,000 blocks at the measured ~143 blocks/s) and, after the faucet claim, a registration confirmation poll (about 1–3 minutes, capped at 300 seconds). Both are expected.
+:::
+
+1. In the sidebar, click **RLN Membership**, then click **Get started** on the welcome screen.
```
weboko, line 65: **A required step is missing here.** Clicking **Get started** does not go to the progress screen — it goes to a **Choose a password** screen with **Password** / **Confirm password** fields and a **Continue** button (subtitle: *"It protects your account on this device."*). **Continue** stays greyed out until both fields match. The wizard only starts provisioning the wallet after that.

```suggestion
1. In the sidebar, click **RLN Membership**, then click **Get started** on the welcome screen.

1. On **Choose a password**, enter a password in both fields and click **Continue**.

   - The password protects the auto-provisioned wallet on this device. **Continue** stays disabled until both entries match.
```

weboko (commented):
Re-dogfooded end-to-end on Linux x86_64 with **Logos Basecamp v0.2.3** against the new `logos-rln-modules` registry, following the doc top to bottom on a clean profile.

## The blocker from my last review is fixed ✅

Registration now completes end-to-end. I landed on **"You're in!"** with membership `pure-gilt-puffin`, rate limit 300 msg/epoch, leaf index 6, registration **Confirmed** at 2026-08-17 19:25:46 UTC.

The old failure (`Cannot resolve dependencies for: logos_execution_zone`) is gone because the new registry replaces that phantom dep with a real, shipped `lez_core` v0.4.1:

| | old `logos-lez-rln` | new `logos-rln-modules` |
|:--|:--|:--|
| `rln_membership_ui` | v0.7.0, dep on `logos_execution_zone` (not in any repo) | v0.7.2, dep on `lez_core` |
| `liblogos_rln_module` | v1.1.0 → `logos_execution_zone` | v0.4.0 → `liblogos_lez_rln_module` |
| package count | 7 | **8** (adds `lez_core` v0.4.1) |

All six deps resolved and every module loaded cleanly:

```
Module loaded: lez_core
Module loaded: liblogos_lez_rln_module
Module loaded: liblogos_rln_module
Module loaded: libp2p_module
Module loaded: rln_gifter_module
Module loaded: keycard_capture_module
```

The **"eight packages"** count on line 42 is accurate — I counted exactly 8 under the **Logos RLN Membership** group. The line 34 claim that deps come from both sources is also correct: `libp2p_module` v1.0.0 resolves from **Logos Official**, the other five from the added registry.

## What still needs fixing

Step 1's UI labels were flagged in my last review and are unchanged, and dogfooding surfaced three new gaps in Step 1–3. Details inline with suggested changes:

1. **`Add repository` button does not exist** (line 36) — unchanged since last review.
2. **`Install with dependencies` button does not exist** (line 44) — unchanged since last review. Worse, a single **Install** click installs *only the 6 dependencies*; the app itself needs a **second** Install, then a Basecamp restart.
3. **An undocumented `Choose a password` step** sits between **Get started** and the progress screen (line 56).
4. **State is `GRACE_PERIOD`, not `Active`**, right after registering (lines 71, 75).
5. **Sync figures are stale** (lines 53, 85) — full sync took ~90s over ~12k blocks, not 5–6 min over ~50k.

## Note on Step 3 explorer verification (no doc change requested)

I could not complete the explorer lookup: the indexer is running well behind the chain. At 19:40 UTC the explorer's newest indexed block was **11867 (17:17:25 UTC)** while the chain head was **~11994**, so my 19:25 tx was still `Transaction not found` 20+ minutes on. The doc's "Indexing may take some time" hedge is correct, so I'm not asking for a change — flagging it because it makes Step 3 unverifiable for a reader following along in real time, and you may want to say the lag can be hours rather than minutes.

Two small things that did work as intended: both **Copy** buttons return the full untruncated values, and the explorer route wants the **bare hash** — the copied value pastes in directly (a `0x`-prefixed hash returns `Invalid transaction hash`).

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,89 @@
...
+
+   - The app lands on **"You're in!"** with your new membership shown below once registration completes.
+
+## Step 3: Verify the membership
+
+Open the membership's detail view to confirm its state, then look up the registration transaction in the block explorer.
+
+1. Tap the membership pill to open its detail view.
+
+   Expect to see state **Active**, a leaf index, your copyable membership id, and the registration transaction.
```
weboko: **The state is `GRACE_PERIOD` immediately after registering, not `Active`.** On the detail view for my brand-new membership the badge read **GRACE_PERIOD**, with **Registration → Status: Confirmed** below it. A reader who follows this line will think something went wrong.

Everything else on this line checks out: leaf index (`6`), a copyable membership id, and the registration transaction were all present, and both **Copy** buttons return full untruncated values.

```suggestion
   Expect to see a leaf index, your copyable membership id, and the registration transaction, with **Registration** showing status **Confirmed**.

   - A newly registered membership shows state **GRACE_PERIOD**, not **Active**.
```

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. Copy the transaction hash and look it up at [the testnet explorer](https://explorer.testnet.lez.logos.co/).
+
+   - Indexing may take some time. Expect membership state **Active**, a leaf index assigned, and rate limit `300`, with the registration transaction visible in the detail view.
```
weboko: Same `Active` correction as above — a freshly registered membership reads **GRACE_PERIOD**.

Rate limit `300` is confirmed (shown as `300 msg/epoch` both on the membership pill and in the detail view).

```suggestion
   - Indexing may take some time — the explorer can run well behind the chain head, so a just-registered transaction may not resolve for a while. Expect membership state **GRACE_PERIOD**, a leaf index assigned, and rate limit `300`, with the registration transaction visible in the detail view.
```

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,89 @@
...
+1. Find **RLN Membership** in the list and click **Install**, then click **Install with dependencies**.
+
+   - **RLN Membership** appears in the sidebar once the install completes.
+
+## Step 2: Register a membership
+
+Open the RLN Membership app and complete the guided wizard to register a new membership.
+
+:::info
+The **Registering** screen runs two unattended waits: a first-run chain sync (about 5–6 minutes for a fresh wallet, roughly 50,000 blocks at the measured ~143 blocks/s) and, after the faucet claim, a registration confirmation poll (about 1–3 minutes, capped at 300 seconds). Both are expected.
```
weboko: **These sync figures no longer match the testnet.** A fresh wallet synced to the chain head in **~90 seconds**, not 5–6 minutes — the head was at block **~11,994**, so the "roughly 50,000 blocks" figure is stale (the testnet looks to have been reset, which fits the registry move and the `liblogos_lez_rln_module` v2.0.0 major bump).

The second wait is accurate: registration confirmed about a minute after the faucet claim, well inside the 300-second cap.

Suggested wording that won't go stale again as the chain grows:

```suggestion
The **Registering** screen runs two unattended waits: a first-run chain sync (a few minutes at most for a fresh wallet, depending on how far the testnet chain head has advanced) and, after the faucet claim, a registration confirmation poll (about 1–3 minutes, capped at 300 seconds). Both are expected.
```

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Frequently asked questions
+
+### Why does the wizard stall at the start of Syncing?
+
+The sequencer was briefly unreachable when the wizard checked for the chain head. Retry, relaunching the app if needed — the wizard resumes from where it left off, and the second attempt typically succeeds.
+
+### Why does the Registering screen take several minutes?
+
+A fresh wallet needs a first-run chain sync, measured at roughly 143 blocks per second over about 50,000 blocks, which takes 5–6 minutes. After the faucet claim, the app polls for registration confirmation for up to 300 seconds. Both waits are expected and the progress bar advances as they proceed.
```
weboko: Same stale figures as line 53 — the measured sync was ~90s to a head of ~11,994 blocks, so "roughly 143 blocks per second over about 50,000 blocks, which takes 5–6 minutes" overstates it by several times and will drift again as the chain grows.

```suggestion
A fresh wallet needs a first-run chain sync, which takes anywhere from under a minute to a few minutes depending on how far the testnet chain head has advanced. After the faucet claim, the app polls for registration confirmation for up to 300 seconds. Both waits are expected and the progress bar advances as they proceed.
```

weboko (commented):
Re-dogfooded end-to-end on a clean Ubuntu 24.04 container with **Logos Basecamp v0.2.3** (AppImage from the `0.2.3` release), following the doc top to bottom on a fresh profile, because `974611c` and `cc56b1f` landed after my last review.

## Result: I completed the tutorial, but only by recovering a URL the doc no longer contains

Registration succeeded — I landed on **"You're in!"** with membership `lunar-silver-antelope`, rate limit `300 msg/epoch`, leaf index `7`, registration **Confirmed** at 2026-08-18 07:19:28 UTC, tx `a2de8a9b442696c247182abf01c54e842bb31bee07b3b0a1877c9cd06c350cf4`.

**Four of the five fixes from my last review are correct and verified:**

| Last review | Status this run |
|:--|:--|
| `catalog` → `catalogue` | ✅ applied |
| `Add repository` → **Manage Repositories** | ✅ applied and accurate — the control is top-right and lands on **Settings › Package Repositories** |
| Missing **Choose a password** step | ✅ applied and accurate — **Continue** stays disabled until both fields match |
| Stale sync figures | ✅ applied — a fresh wallet synced blocks 7000→12706 in **15 seconds**, so "a few minutes at most" is a safe upper bound |
| `Active` → `GRACE_PERIOD` | ❌ not applied — **and it should not be.** See the note on line 82. |

## What needs fixing

1. ⛔ **`974611c` deleted the registry URL** (line 42). This is a hard blocker: Step 1 is now unfollowable. Details and a restore suggestion inline.
2. **The two-pass install and the mandatory restart do not happen on v0.2.3** (lines 46–53) — one **Install** click installed all six dependencies *and* the app, and the sidebar entry appeared without a restart.
3. **Step 3's explorer lookup is not completable in real time** (lines 84–86) — the indexer advanced ~2 blocks in 10 minutes while my transaction sits above its head.
4. **The version callout says v0.2.1** (line 18) but the software this doc drives is Basecamp 0.2.3.

## On the `GRACE_PERIOD` suggestion I left last time — please don't apply it

My brand-new membership read **ACTIVE** (green badge) the first time I opened the detail view, not `GRACE_PERIOD`. The doc is right as written on lines 82 and 86; applying my earlier suggestion would introduce an error. I've left a note there.

## Note on the version callout added in `cc56b1f`

`cc56b1f` adds the same `:::tip[Version]` block to **41 files** across storage, messaging, blockchain, mixnet, and core — none of which this PR otherwise touches. That's a repo-wide convention change riding inside an RLN feature PR, and the number it introduces looks wrong (see line 18). Worth splitting into its own PR so the convention and the number can be reviewed on their own; as it stands, one wrong string is now in 41 places.

## Things I verified that are correct and need no change

- The registry URL resolves and its index lists exactly **eight** packages under **Logos RLN Membership**.
- `libp2p_module` v1.0.0 resolves from **Logos Official**; the other five come from the added registry — line 48's claim holds.
- All six modules load cleanly on restart (`lez_core`, `liblogos_lez_rln_module`, `liblogos_rln_module`, `libp2p_module`, `rln_gifter_module`, `keycard_capture_module`).
- The three progress stages really are **Syncing** → **Claiming** → **Registering** ("Syncing with Logos Blockchain...", "Claiming faucet tokens...", "Registering membership...").
- The faucet claim completed without ever showing **Get more tokens**, and registration confirmed inside the documented 300-second budget.
- Both **Copy** buttons return full untruncated values, and the transaction hash comes back bare (no `0x`).
- The ~1 GB disk prerequisite is about right: 278 MB AppImage + ~300 MB of module downloads, 216 MB installed.

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,100 @@
...
+slug: register-rln-membership-from-basecamp
+---
+
+# Register an RLN membership from Basecamp
+
+#### Get started acquiring an RLN membership on the LEZ testnet.
+
+:::tip[Version]
+This document is accurate for **Testnet v0.2.1**.
+:::
```
weboko, line 19: **This version number doesn't match anything the reader will be running.**

`cc56b1f` introduces this callout here and in 40 other files, but v0.2.1 isn't the version of the software this tutorial drives:

- Basecamp's latest release is **0.2.3** (the app itself prints `0.2.3` in the sidebar), and that's what I dogfooded on.
- The RLN modules are versioned independently and don't track a "0.2.x" line at all: `rln_membership_ui` v0.7.2, `lez_core` v0.4.1, `liblogos_rln_module` v0.4.0.
- Everywhere else the repo talks about the testnet it says **"testnet v0.2"**, not v0.2.1 — see `docs/lez/programs/trade-tokens-on-lez-amm-program.md:16` and `docs/lez/accounts/set-up-shared-private-lez-account.md:18`.

The nearest thing to a "v0.2.1" is the Blockchain **UI module** version pinned in `docs/blockchain/get-started/run-a-logos-blockchain-node-from-basecamp.md:51`, which is a different component entirely.

Naming what was actually verified is more useful to a reader than a release train number:

```suggestion
:::tip[Version]
This document is accurate for **Logos Basecamp v0.2.3** with **RLN Membership v0.7.2**.
:::
```

The same correction applies to the other 40 files this commit touches, whichever number you settle on.

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Step 1: Install the RLN Membership app
+
+Add the RLN module registry, then install the app with its dependencies from both the registry and the official catalogue.
+
+1. In Basecamp's **Package Manager**, click **Manage Repositories** (top-right) to open **Settings › Package Repositories**.
+
+1. Under **Add a repository**, paste the registry URL and click **Add**:
+
+
```
weboko: ⛔ **Blocker: the registry URL is gone, so Step 1 cannot be completed.**

`974611c` restructured this step to use **Manage Repositories** (correct — that was my suggestion) but dropped the fenced block that carried the URL along with it. Line 42 tells the reader to "paste the registry URL" and line 44 is an empty indented line. There is nothing to paste, and no other step supplies the value.

This is not theoretical — it's exactly where I stalled. **Settings › Package Repositories** opens with an empty **Add a repository** field and only the built-in **Logos Official** entry; without the URL there is no way to reach the RLN packages, and every remaining step depends on it.

I recovered the URL from `2512297` (the commit before the restructure) and it is still live — `HTTP 200`, and adding it registered **Logos RLN Membership** and let me finish the whole tutorial. Its index resolves to exactly **eight** packages, so the verification line that `974611c` also removed was accurate and is worth restoring alongside it:

````suggestion
1. Under **Add a repository**, paste the registry URL and click **Add**:

   ```text
   https://github.com/logos-co/logos-rln-modules/releases/download/index/logos-repo.json
   ```

   - The repository lists as **Logos RLN Membership** with eight packages.
````

File: docs/lez/rln/register-rln-membership-from-basecamp.md [hunk 14]
```diff
@@ -0,0 +1,100 @@
...
+   
+
+1. Find **RLN Membership** in the list and click **Install**. In the **Install Package?** dialog, review the listed dependency changes and click **Install** to confirm.
+
+   - The wallet, libp2p, and RLN modules are installed for you — `libp2p_module` resolves from the official catalogue and the rest from the registry you added.
+   - This first pass installs the dependencies only. When it finishes, click **Install** on **RLN Membership** again to install the app itself; the dialog confirms **No other packages need to change**.
+
+1. Restart Basecamp.
+
+   - **RLN Membership** appears in the sidebar after the restart. If the row shows **RETRY** before you restart, the install has still succeeded — restart and check the sidebar before retrying.
```
weboko: **The two-pass install and the mandatory restart didn't happen on v0.2.3 — one Install click did everything.**

I clicked **Install** once and confirmed the dialog once. The package log shows all six dependencies *and* the app downloading in that single pass:

```
07:15:55  keycard_capture_module-0.2.0.lgx
07:15:57  rln_gifter_module-0.2.0.lgx
07:15:58  libp2p_module-1.0.0.lgx
07:16:00  liblogos_lez_rln_module-2.0.0.lgx
07:16:02  liblogos_rln_module-0.4.0.lgx
07:16:11  lez_core-0.4.1.lgx
07:16:23  rln_membership_ui-0.7.2.lgx   <-- the app, same pass
```

The **RLN Membership** row went straight from `INSTALLING...` to `INSTALLED`, so there was no second **Install** button to click — following line 49 literally leaves you hunting for a control that isn't there. And **RLN Membership** appeared in the sidebar *before* I restarted, so line 53's "after the restart" isn't required either (I confirmed the tooltip on the new sidebar icon while the old process was still running).

I think this is what my last review actually caught: the app is the *last* of the seven downloads and lands ~30 seconds after the first dependency, so checking the row too early shows the deps installed and the app not yet. The robust instruction is "wait for INSTALLED", which covers both.

```suggestion
1. Find **RLN Membership** in the list and click **Install**. In the **Install Package?** dialog, review the listed dependency changes and click **Install** to confirm.

   - Six dependencies are installed for you — `libp2p_module` resolves from the official catalogue, and `lez_core`, `liblogos_lez_rln_module`, `liblogos_rln_module`, `rln_gifter_module`, and `keycard_capture_module` from the registry you added.
   - The single confirmation installs the dependencies and the app itself, downloading about 300 MB. Wait for the **RLN Membership** row to read **INSTALLED** before moving on; if it still offers **INSTALL**, click it once more.

1. Open **RLN Membership** from the sidebar, restarting Basecamp first if it isn't there yet.

   - If the row shows **RETRY**, the install has still succeeded — restart and check the sidebar before retrying.
```

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,100 @@
...
+
+   - The app lands on **"You're in!"** with your new membership shown below once registration completes.
+
+## Step 3: Verify the membership
+
+Open the membership's detail view to confirm its state, then look up the registration transaction in the block explorer.
+
+1. Tap the membership pill to open its detail view.
+
+   Expect to see state **Active**, a leaf index, your copyable membership id, and the registration transaction.
```
weboko: **This line is correct — please don't apply my earlier `GRACE_PERIOD` suggestion.**

I flagged this last review and I was wrong. On this run the detail view for a brand-new membership showed a green **ACTIVE** badge the first time I opened it, about a minute after registering:

```
lunar-silver-antelope
  [ACTIVE]
  Rate limit     300 msg/epoch
  Leaf index     7
  Membership id  baa315ac4e...3cfa9b05   [Copy]
  Registration
  Status         Confirmed
  Transaction    a2de8a9b44...6c350cf4   [Copy]
  Registered     2026-08-18 07:19:28 UTC
```

Everything else on this line checks out too: leaf index (`7`), a copyable membership id, and the registration transaction were all present. `GRACE_PERIOD` is presumably a transient state I happened to catch last time; **Active** is the right thing to document as the expected outcome. No change needed here, and the unresolved suggestion above it should be dismissed rather than applied.

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. Copy the transaction hash and look it up at [the testnet explorer](https://explorer.testnet.lez.logos.co/).
+
+   - Indexing may take some time. Expect membership state **Active**, a leaf index assigned, and rate limit `300`, with the registration transaction visible in the detail view.
```
weboko: **This step still can't be completed by a reader following along, and "may take some time" undersells it by a lot.**

I measured the lag this time rather than just hitting the error:

- My transaction is definitely on chain — the sequencer's `getTransaction` RPC returns it.
- The explorer returned `Transaction not found` for the full 15 minutes I polled it, in 30-second intervals.
- Its indexed head was block **12669** when I registered and **12671** ten minutes later — about **2 blocks per 10 minutes**, while my node had already synced to **12706**. The indexer is falling further behind, not catching up, so at that rate this transaction is hours away from being visible.

There's also a navigation gap: the link goes to the explorer root, but the root has no obvious search affordance for this, and the working route is `/transaction/<hash>` (`/tx/<hash>` 404s). Worth naming it so the reader isn't hunting.

```suggestion
1. Copy the transaction hash and look it up at [the testnet explorer](https://explorer.testnet.lez.logos.co/), whose transaction route is `https://explorer.testnet.lez.logos.co/transaction/<hash>`.

   - Paste the hash exactly as **Copy** returns it. The explorer expects the bare hash; a `0x` prefix returns `Invalid transaction hash`.
   - The explorer's indexer runs well behind the chain head, so a just-registered transaction can return `Transaction not found` for hours. Your registration is already confirmed in the app, so treat this lookup as optional later confirmation rather than a step to wait on. In the app's detail view, expect membership state **Active**, a leaf index assigned, and rate limit `300`.
```

docs/lez/rln/register-rln-membership-from-basecamp.md
kashepavadan: @weboko Looks like a correct fix in the github view but just erases it in reality. Second time this is happening.

kashepavadan:
@weboko just need a final sanity check please

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,100 @@
...
+steps_layout: sectioned
+authors: adklempner, kashepavadan
+owner: logos
+doc_version: 1
+slug: register-rln-membership-from-basecamp
+---
+
+# Register an RLN membership from Basecamp
+
+#### Get started acquiring an RLN membership on the LEZ testnet.
```
weboko, line 15: **Please restore the version callout — and I owe you a retraction here, because my last review is why it went away.**

I flagged the callout `cc56b1f` added, on the grounds that "**Testnet v0.2.1**" didn't match anything the reader would be running (I'd just installed Basecamp **0.2.3**). `f0d0aec` then reverted it entirely. **My objection was wrong, and reverting overshot** — I was reading the callout as naming the *Basecamp build*, when the repo convention is that it names the ***Testnet release***. Those are different things, and `v0.2.1` is the correct value.

What settles it is that `#457` merged into `main` today (it's already in this branch via the `bbaa0e44` merge) and added this exact callout to **39 pages** — every single `doc_type: procedure` doc in the repo, all with the identical string:

```
:::tip[Version]
This document is accurate for **Testnet v0.2.1**.
:::
```

That includes `docs/basecamp/install-logos-basecamp.md` — the prerequisite **this page links to on line 21** — and all 10 other `docs/lez/**` procedure pages.

After the revert, this page is now **the only `doc_type: procedure` doc in the repo without one**. The sole other `docs/lez` page missing it is `introduction-to-the-logos-execution-zone.md`, which is `doc_type: concept` and therefore genuinely exempt. So this is a real gap, not a style preference.

Restoring it in the sibling position (directly under the `####` subtitle, matching all 39):

```suggestion
#### Get started acquiring an RLN membership on the LEZ testnet.

:::tip[Version]
This document is accurate for **Testnet v0.2.1**.
:::
```

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,100 @@
...
+1. In Basecamp's **Package Manager**, click **Manage Repositories** (top-right) to open **Settings › Package Repositories**.
+
+1. Under **Add a repository**, paste the registry URL and click **Add**:
+
+   ```text
+   https://github.com/logos-co/logos-rln-modules/releases/download/index/logos-repo.json
+   ```
+   - The repository lists as **Logos RLN Membership** with eight packages.
+
+1. Find **RLN Membership** in the list and click **Install**. In the **Install Package?** dialog, review the listed dependency changes and click **Install** to confirm.
```
weboko: `dialog` trips `Logos.BritishSpelling`, which is **`level: error`** in `.github/styles/Logos/BritishSpelling.yml` — the same house rule behind the `catalog` → `catalogue` fix you already applied on line 34.

Worth fixing rather than ignoring: across the merged docs this rule has only **4** hits in total, so it's one the repo actually keeps clean. In fairness there is one existing precedent that also says "dialog" (`docs/mixnet/get-started/send-anonymised-messages-over-the-mix-network.md:98`), so this is a nit, not a defect — but the wording is otherwise correct and I confirmed the dialogue is titled exactly **Install Package?**.

```suggestion
1. Find **RLN Membership** in the list and click **Install**. In the **Install Package?** dialogue, review the listed dependency changes and click **Install** to confirm.
```

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/lez/rln/register-rln-membership-from-basecamp.md
```diff
@@ -0,0 +1,100 @@
...
+   - If the wizard stalls at the start of **Syncing**, or shows "Cannot discover the chain head — is the sequencer reachable?", retry — relaunch the app if needed. The wizard resumes where it left off, and the second attempt typically proceeds.
+   - If **Get more tokens** appears, the faucet claim timed out after its 180-second budget. Click the offered button to retry.
+
+   - The app lands on **"You're in!"** with your new membership shown below once registration completes.
+
+## Step 3: Verify the membership
+
+Open the membership's detail view to confirm its state, then look up the registration transaction in the block explorer.
+
+1. Tap the membership pill to open its detail view.
```
weboko: **"Tap" is a touch gesture, and this is a desktop app.**

This is the only instance of "Tap" in the entire `docs/` corpus — I grepped every merged Markdown file and it appears nowhere else. This same page uses "click" **10 times** (lines 36, 38, 45, 61, 63, 70…), so it's internally inconsistent too. I did the action with a mouse click on Basecamp 0.2.3 and it opened the detail view as described.

```suggestion
1. Click the membership pill to open its detail view.
```

---
_Generated by [Claude Code](https://claude.ai/code)_

weboko (commented):
Re-dogfooded end-to-end on a clean Ubuntu 24.04 container with **Logos Basecamp v0.2.3** (AppImage from the `0.2.3` release) on a fresh profile, at `bbaa0e44` — because `f0d0aec` (revert of the version callout), `e817dc5` (applied suggestions) and `5b805ea` landed after my last review and materially rewrote Step 1.

## Result: the procedure now works end-to-end, exactly as written ✅

This is the first run where I completed the tutorial **without having to recover anything the doc was missing.** Registered `open-peridot-quokka` — state **ACTIVE**, rate limit `300 msg/epoch`, leaf index `8`, registration **Confirmed**, tx `a56980fcd8e62051888f89b1eedf0fc35698a4a6f2db4ff6bac15342c42b226a` at 2026-08-18 19:21:10 UTC.

**Every fix from my last review is applied and verified correct:**

| Last review | Status this run |
|:--|:--|
| ⛔ `974611c` deleted the registry URL | ✅ restored in a `text` code block, and it works — pasting it added the repo |
| Two-pass install + mandatory restart don't happen on v0.2.3 | ✅ corrected, and verified: **one** Install confirmation installed all 6 deps *and* the app, and the app appeared in the sidebar **with no restart** |
| Explorer step unfollowable / "may take some time" undersells it | ✅ fixed — and both new claims are exactly right (see below) |
| `Active` (my earlier `GRACE_PERIOD` suggestion) | ✅ correctly **not** applied — detail view reads `ACTIVE` immediately |

**Claims I actively tested rather than eyeballed:**

- **Line 43, "eight packages"** — accurate. The registry index lists exactly 8 (`eth_auth_module`, `keycard_auth_module`, `keycard_capture_module`, `lez_core`, `liblogos_lez_rln_module`, `liblogos_rln_module`, `rln_gifter_module`, `rln_membership_ui`), and it does list as **Logos RLN Membership**.
- **Line 34, deps "from both the registry and the official catalogue"** — accurate. The dialogue showed `libp2p_module v1.0.0` sourced from **Logos Official** and the other five from the added registry.
- **Line 47, "about 300 MB"** — sane. Package sizes sum to ~300 MB (`lez_core` alone is ~210 MB).
- **Line 65, Continue stays disabled until both entries match** — verified: with only the first field filled, **Continue** is greyed out; it enables on the matching second entry.
- **Line 58, the two unattended waits** — both well inside the stated bounds. Fresh-wallet sync took **~60 s** (to block 13426), the faucet claim ~40 s, and registration confirmation ~65 s. Total Continue → "You're in!" was **2 m 50 s**.
- **Line 82/84/85, the explorer** — this is the part I most expected to be wrong, and it is **exactly right**. `Copy` returns a bare hash with no `0x`; `…/transaction/<bare-hash>` returns `Transaction not found` (indexer lag, as documented), and `…/transaction/0x<hash>` returns **`Invalid transaction hash`**, the precise string on line 84.
- **Build** — `npm run build` passes, so the relative prerequisite link resolves under `onBrokenLinks: 'throw'`.

## What still needs fixing

Three small things, none of them blocking. The first is the one that matters — it's a regression the revert introduced against a convention that landed on `main` *today*.

1. **The version callout should come back** (line 15). Details and a suggestion inline — and please note I am **retracting my own earlier objection** to it, with an explanation, so this doesn't get whipsawed a third time.
2. `dialog` → `dialogue` (line 45), `Logos.BritishSpelling`, level `error`.
3. `Tap` → `Click` (line 78) — a desktop app, and the only "Tap" in the whole docs corpus.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #429, opened 2026-08-05
docs(messaging): update to chat_module v0.2.2

igor-sirotin:
Updates the chat-module procedure to **chat_module v0.2.2**, built against **delivery_module v0.2.0** (no chat_module 0.1.x works with delivery 0.2.0).

API changes reflected in the doc:

- `init()` takes a single `ChatConfig` record (`delivery_preset`, `log_level`; a `QVariantMap` from C++) — no more instance path or TCP port.
- Intro bundles are gone: peers exchange the reusable address from `get_address()`, and `create_conversation(peer_address)` takes just the address.
- Events: `message_received` gained `sender`, `conversation_created` gained `kind`/`name`/`desc`, new `members_changed`.
- Persistence is disabled in this release — the persistence notes now describe ephemeral state.
- Flake wiring mirrors logos-chat-ui v0.2.2; links pinned to v0.2.2.

Verified against `chat_module.lidl` @ v0.2.2 and logos-chat-ui v0.2.2 (`ChatBackend.cpp`, `flake.nix`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #431, opened 2026-08-08
docs(messaging): add group conversations to the chat-module procedure

igor-sirotin:
Extends `build-logos-module-that-uses-chat-module-api.md` (on top of #429) with group chat support, covering journey [logos-co/journeys.logos.co#26](https://github.com/logos-co/journeys.logos.co/issues/26). Same instruction style as the existing direct-chat flow; group encryption is de-MLS (there is no Sender Keys fallback in the stack).

- Step 4 gains group lifecycle items: `create_group_conversation(name, desc)`, `add_group_member` with its asynchronous commit semantics (proposal → steward commit after the 60 s inactivity window → welcome; joins land minutes later), `list_group_members` with the `pending` flag, and group send fan-out with sender attribution (one send per message regardless of group size; brief post-commit windows reject sends).
- New Step 7: a three-instance verification walkthrough mirroring `doctests/chat-module-group.test.yaml` in logos-chat-module v0.2.2, including a non-creator add.
- Two new troubleshooting entries: invited member not joined yet (async by design), send rejected right after a membership change.
- Removes two stray `:::` closers left in the base branch.

All names and semantics verified against `rust-lib/chat_module.lidl` @ v0.2.2 and the group doctest (Doc-Tests CI green at the tag).


igor-sirotin:
@kashepavadan this is ready for review. Please let me know if you think it's better to have this journey as a separate doc. But as you can see, quite big parts would be duplicated (setup, initialization, shutdown, testing, etc).

docs/messaging/chat-module/build-logos-module-that-uses-chat-module-api.md
```diff
@@ -303,6 +353,20 @@ A chat is only proven end to end when a message travels between two running inst
...
 Seeing the `message_received` events on both sides confirms the full round trip: identity, address exchange, key discovery, conversation setup, and end-to-end encrypted delivery.
 
+## Step 10: Verify a three-instance group conversation
+
+Group semantics only show with three or more members. Start three instances (A, B, and C) the same way as in Step 9 and wait for all three to report `online`. Group joins land asynchronously (see Step 6) — budget minutes per membership change.
+
+1. Collect each instance's address via `get_address()`.
+1. In instance A, call `create_group_conversation("Book Club", "Weekly sci-fi picks")` and keep the returned conversation id.
+1. In instance A, call `add_group_member(groupId, addressOfB)`. The call returns immediately; B joins once the group commits the add — B receives a `conversation_created` event (`kind == "group"`, name `"Book Club"`) under the same conversation id A holds.
+1. In instance B, call `add_group_member(groupId, addressOfC)` (any member can grow the group). C joins the same way.
```
weboko, line 363: For B to invite C, B must already be a **committed** member of the group: `add_group_member` from B can only succeed once B has actually joined (it needs to hold the group's current MLS epoch). The upstream group doctest makes this an explicit *"Bob joins"* poll **between** the two invites; the step here jumps straight from "A adds B" to "B adds C", which a reader driving three instances by hand can hit before B's join has committed. Suggest making the ordering explicit:

```suggestion
1. Once B has joined (the `conversation_created` from the previous step has arrived on B), call `add_group_member(groupId, addressOfC)` from instance B — any member can grow the group. C joins the same way.
```

weboko (commented):
## Dogfooded end-to-end ✅

I followed this procedure from a clean machine and verified the group-conversation additions against the shipped `chat_module` **v0.2.2** contract.

**What I ran**
- Installed Nix (flakes enabled) and scaffolded per Step 1: `nix flake init -t github:logos-co/logos-module-builder/tutorial-v3#ui-qml-backend` → the `ui_example` template with exactly the files the Step 1 table lists.
- Applied Step 2 verbatim (`metadata.json` deps + `dependency_overrides`; `flake.nix` with `logos-module-builder.follows = "chat_module/logos-module-builder"`, `chat_module` pinned to `v0.2.2`, `logos-delivery-module` to `v0.2.0`). `nix flake metadata` resolves the whole input graph cleanly — the `follows`/`flakeInputs` wiring is correct.
- Wired the Step 3–7 code into `UiExamplePlugin::initLogos` — the event subscriptions plus `init`, `get_address`, `create_conversation`, `create_group_conversation("Book Club", "Weekly sci-fi picks")`, `add_group_member`, `send_message`, `list_conversations`, `get_messages`, `list_group_members` — and ran Step 8's `nix build`. **It builds and links** (`ui_example_plugin.so`), so every documented method name/signature and all seven events (incl. `members_changed`) match the generated typed client.

**Contract cross-check** — every group claim matches `rust-lib/chat_module.lidl` @ v0.2.2 and the repo's own `doctests/chat-module-group.test.yaml`:
- `create_group_conversation(name, desc)`, `add_group_member(convo_id, peer_address)`, `list_group_members → [GroupMember]{address, pending}` ✓
- `conversation_created(… kind, name, desc)`, `members_changed`, and no member-removal method ✓
- async commit semantics + "commit-inactivity window (60s default), join lands minutes later" ✓ (matches the doctest's de-mls timing note)

The live Steps 9–10 message round trip needs the real `logos.test` Waku fleet — the upstream doctest itself notes *"an isolated sandbox won't complete the exchange"* — so I couldn't run the three-instance exchange in this sandbox. The build + contract + doctest cross-checks all pass.

One small clarity suggestion on Step 10 inline.

weboko:
human review LGTM too

PR #432, opened 2026-08-08
docs(messaging): update the delivery module journey to v0.2.0

igor-sirotin:
Brings [Use the Logos Delivery module API from an app](https://docs.logos.co/messaging/delivery-module/use-logos-delivery-module-api-from-app) up to date with the `delivery_module` **v0.2.0** release. 

Main change — `createNode` method arguments.


weboko (approved):
## Reviewed + verified against the reference demo ✅

Accurate v0.1.3 → v0.2.0 migration. All referenced pins resolve (delivery-module `v0.2.0` = commit `3258cdb`, `logos-module-builder` `0.2.5` has the `ui-qml-backend` template, `logos-delivery-demo` `v0.2.0` exists), and — more importantly — every substantive claim matches the actual working code of the reference demo the page points to (`logos-delivery-demo` @ v0.2.0):

- **Layered `createNode` config** — `{ "mode": "Core", "preset": "logos.test" }` matches the delivery-module v0.2.0 README verbatim; `mode` = Core(relay)/Edge(light) and the `messagingOverrides`/`channelsOverrides` layering are confirmed there and in `delivery_module_plugin.h`.
- **`logLevel` under `messagingOverrides`** — the demo plugin builds exactly `{{"mode",…},{"preset",…},{"messagingOverrides", {{"logLevel","INFO"}}}}`, so the troubleshooting change to `"messagingOverrides": {"logLevel": "DEBUG"}` is right.
- **Port behaviour** — the demo comment states it directly: *"the layered shape gets ephemeral p2p ports (logos-delivery defaults them to 0)… one bare field switches parsing to the legacy flat path, whose fixed port defaults do collide."* That's exactly your "any other top-level key selects the legacy flat shape" warning and the two-instances-collide entry.
- **Builder ≥ 0.2.5 + `follows` wiring** — the demo's `flake.nix` is line-for-line what you document (`logos-module-builder` pinned to `0.2.5`, `delivery_module.inputs.logos-module-builder.follows = "logos-module-builder"`), with the same rationale: *"Below 0.2.5 binary event payloads arrive empty (logos-cpp-sdk#99)."*
- The flat shape *"still parses and boots the full stack"* per the README, matching how you frame it as legacy-but-valid.

No changes requested. (I confirmed the module-builder toolchain builds a QML module in a separate chat-module build this run; the delivery flake/config here is identical in shape to that working demo, so I didn't re-run the full node build — the live `start()`/online path needs the `logos.test` Waku fleet regardless.)

weboko:
fyi there are some comments from bot, but to see them you need to open diff 

image

kashepavadan (approved):

PR #433, opened 2026-08-08
docs(messaging): add the Reliable Channel API to the delivery module journey

igor-sirotin:
Covers journey [logos-co/journeys.logos.co#23](https://github.com/logos-co/journeys.logos.co/issues/23) — "Use the Reliable Channel API from an app".

> **Stacked on #432** (the v0.2.0 update), which must merge first — the Reliable Channel API does not exist before `delivery_module` v0.2.0. Review this PR's own commits; the base diff belongs to #432.

Rather than a separate page, this adds the journey to [Use the Logos Delivery module API from an app](https://docs.logos.co/messaging/delivery-module/use-logos-delivery-module-api-from-app) and reshapes the page the same way #431 reshaped the chat-module procedure: shared setup and shutdown wrapped around **one section per API**.

## New structure

| Step | |
|---|---|
| 1–2 | Scaffold the module, declare the dependency *(unchanged)* |
| 3 | Initialise `LogosModules` and subscribe to events — messaging **and** channel handlers, before the node starts |
| 4 | Create and start the node |
| **5** | **Use the Messaging API** — subscribe, send, receive, unsubscribe |
| **6** | **Use the Reliable Channels API** — create, exists, send, receive, close, plus `channelsOverrides` |
| 7 | Shut down |
| 8 | Build and run |
| 9 | Verify a two-instance message exchange |
| 10 | Verify a two-instance reliable channel exchange |


docs/messaging/delivery/use-logos-delivery-module-api-from-app.md
```diff
@@ -187,14 +214,99 @@ Register event handlers before calling `start()` so you don't miss the first `co
...
+       qWarning() << "channelCreate failed:" << r.getError();
+       return;
+   }
+   ```
+
+   Creating a channel subscribes its content topic for you — you don't call `subscribe()` for a channel topic.
+
+   :::caution
+   Do not re-create a channel id you have already closed: `channelCreate` with the id of a closed channel currently crashes the delivery node process. Use a fresh channel id instead, and keep a channel open for as long as your application needs it.
+   :::
```
weboko, line 250: Good to document this, but it **directly contradicts** the upstream v0.2.0 header: `src/delivery_module_plugin.h` documents `channelCreate` as *"Creates (or re-opens) a reliable channel. Persisted channel state survives `channelClose`, so re-creating a channel with the same id restores it."* — i.e. exactly the post-close re-open you're warning against, advertised as supported. The two agree on the mechanism (close persists SDS state, create restores it) but disagree on the outcome, so this reads as a real bug in the restore path rather than intended behaviour.

Two small asks so the caveat stays honest and removable:
1. Link the tracking issue for the crash (the null-deref detail in the troubleshooting entry shows you hit it empirically) so readers see it's a known bug, not the contract.
2. Worth flagging to the delivery-module team that the `channelCreate` doc-comment still advertises re-open as supported — otherwise the header and this page stay in conflict.

weboko (commented):
## Reviewed against v0.2.0 source ✅

The Reliable Channels section is accurate against `logos-delivery-module` @ v0.2.0. I checked every method and event in `src/delivery_module_plugin.h` and they match exactly:

- **Methods** — `channelCreate(channelId, contentTopic, senderId)`, `channelExists(channelId)` (returns the verbatim FFI `"true"`/`"false"` string, so `getString() == "true"` is right), `channelSend(channelId, payload)` (returns requestId; results via `channelMessageSent`/`channelMessageError`), `channelClose(channelId)`.
- **Events** — `channelMessageReceived(channelId, senderId, payload, ts)`, `channelMessageSent(channelId, requestId, ts)`, `channelMessageError(channelId, requestId, error, ts)` — the `data[…]` index lists are correct.
- **Kernel-only** — `channel*` failing with `no reliable channel manager`, and `"channels"` being the default `entryLayer`, both match the README / source; the `channel already exists` string is in the source too.
- The empty-payload-on-builder-<0.2.5 caveat is consistent with the same wire-format issue documented in the Messaging section (#432).

Step renumbering (3→10) and cross-refs are internally consistent.

Couldn't verify from source (grammar lives in `logos-messaging/logos-delivery`, outside this session's scope): the `channelsOverrides` key names/defaults (`sdsAcknowledgementTimeoutMs` 5000, `sdsMaxRetransmissions` 5, `sdsCausalHistorySize` 2), the channelCreate auto-subscribe, and the re-open crash — I couldn't reproduce the runtime crash here (needs a node on the live `logos.test` fleet). One inline note on that caveat.

weboko:
on this PR too, not sure why they are not visible on the main page of PR 

image

docs/messaging/delivery/use-logos-delivery-module-api-from-app.md
kashepavadan: @igor-sirotin what's going on here?

kashepavadan (approved):
Depends on solving the contradiction with the code in Sasha's review

docs/messaging/delivery/use-logos-delivery-module-api-from-app.md
igor-sirotin: This is a bug I discovered while writing this doc. Here's the link:
- https://github.com/logos-messaging/logos-delivery/issues/4116

weboko:
@kashepavadan should these be addressed? 
https://github.com/logos-co/logos-docs/pull/433#issuecomment-5234343485

I have same on my other PRs

kashepavadan:
@weboko Ignore for now. I want to do a pass fixing these across all the docs at once at some point later. It's not a priority atm

weboko (commented):
Re-reviewed after the latest commits (per-API split, the e2e corrections, and the [logos-delivery#4116](https://github.com/logos-messaging/logos-delivery/issues/4116) crash link).

I dogfooded the procedure on Linux x86_64: installed Nix with flakes, scaffolded with `nix flake init -t github:logos-co/logos-module-builder/0.2.5#ui-qml-backend`, and `nix build` succeeds, producing `ui_example_plugin.so`. The Reliable Channels surface in Steps 3–10 checks out against the pinned `delivery_module_plugin.h` and the `logos-delivery-demo` reference: `entryLayer` default `channels`, kernel-only `no reliable channel manager`, `channelExists` returning verbatim `"true"`/`"false"`, and the event signatures all match. Two findings:

**1. Blocker — Step 1.4's `rm` breaks the build.** (Pre-existing content, outside this PR's diff, so I couldn't attach an inline suggestion — but it blocks the very journey this PR completes, so flagging it here.) Step 1.4 runs:

```
rm -f src/ui_example.rep src/ui_example_interface.h src/ui_example_plugin.h src/ui_example_plugin.cpp
```

With the `0.2.5#ui-qml-backend` template that Step 1.2 pins, `ui_example_interface.h`, `ui_example_plugin.h`, and `ui_example_plugin.cpp` **do not exist** — that layout is the older/plugin authoring model (it's what the hand-written `logos-delivery-demo` uses). The only file the command actually deletes is `src/ui_example.rep`, which `CMakeLists.txt` references as `REP_FILE`. I verified end to end: the pristine scaffold builds fine as-is, but running Step 1.4 then `nix build` fails with:

```
Error: cannot read .rep file: src/ui_example.rep
```

So the step's premise (*"leaving them causes build errors"*) is inverted — the example sources build, and the `rm` is what breaks the build. **Suggested fix:** drop Step 1.4 for the `#ui-qml-backend` template (you edit/rename `ui_example_backend.{h,cpp}` + `ui_example.rep`; there are no example plugin files to remove). If the intent is instead the plugin model that Steps 3–7 describe (`initLogos()` in a `*_plugin.cpp`, matching the demo), then Step 1.2 needs a plugin template rather than `#ui-qml-backend` — note that 0.2.5 only ships `ui-qml-backend` and a QML-only `ui-qml`, so reconciling the scaffold with the demo-style code is worth a follow-up. (The same stale `rm` lives in `core/build-modules/build-a-logos-cpp-ui-module.md`.)

**2. Timestamp type** — see the inline suggestion on the `messageReceived` handler.

docs/messaging/delivery/use-logos-delivery-module-api-from-app.md
```diff
@@ -133,12 +134,32 @@ Register event handlers before calling `start()` so you don't miss the first `co
        // data[3]: QString     — timestamp (ns since epoch)
```
weboko: **Dogfooding note — timestamp type.** The reference `logos-delivery-demo` (v0.2.0) reads this timestamp as `qint64` — `data.at(3).toLongLong()`, commented *"timestamp (qint64, ns since epoch)"* — and the pinned `delivery_module_plugin.h` (@3258cdb) declares every event `timestamp` as `int64_t`. So `data[3]` arrives as `qint64`, not `QString`. This also disagrees with the `channelMessageReceived` handler added just below (line 147), which correctly types `data[3]` as `qint64`. The `connectionStateChanged` handler above has the same slip on its `data[1]` timestamp (line 127). Suggested fix for this line (the `data[1]` line above wants the same change):

```suggestion
       // data[3]: qint64      — timestamp (ns since epoch)
```

kashepavadan:
@igor-sirotin take a look at the comment above please

weboko (approved):

igor-sirotin:
@weboko thanks for the finding, I'll fix it

igor-sirotin:
@weboko @kashepavadan Thanks again for the review. I've updated the doc to properly work with `logos-module-builder/0.2.5#ui-qml-backend` template. Please check

kashepavadan (approved):

docs/messaging/delivery/use-logos-delivery-module-api-from-app.md
```diff
@@ -186,23 +219,108 @@ Register event handlers before calling `start()` so you don't miss the first `co
 1. Send a message. On success, `getString()` returns the request ID; track it through `messageSent` → `messagePropagated` events (or `messageError`):
```
weboko: The two events are in the wrong order here. The lifecycle is `messagePropagated` (message reached the network) **then** `messageSent` (message confirmed/validated by the network) — propagated comes first. This matches the pinned `delivery_module_plugin.h` @`3258cdb` ("`messagePropagated` emitted if the message has hit the network" / "`messageSent` emitted after the message is validated by the network"), the delivery-module v0.2.0 README (§Events: propagated = "reached the network but is not yet validated", sent = "confirmed by the network"), and this PR's own Step 9 (line 356), which already lists them in that order ("`messagePropagated` and `messageSent`").

```suggestion
1. Send a message. On success, `getString()` returns the request ID; track it through `messagePropagated` → `messageSent` events (or `messageError`):
```

weboko (commented):
Dogfooded this end-to-end on a clean Ubuntu 24.04 / x86_64 box with a fresh Nix install (flakes), following the doc verbatim against the pinned `0.2.5` template and `delivery_module` `v0.2.0`.

**Confirmed working**
- **Step 1** scaffold from `github:logos-co/logos-module-builder/0.2.5#ui-qml-backend`, and “the scaffold builds as it stands” — bare `nix build` succeeds.
- **Step 2** dependency wiring (the `flake.nix` `0.2.5` pin + `follows`, and `metadata.json` `"dependencies": ["delivery_module"]`).
- **Steps 3–7** compile exactly as written against module-builder `0.2.5`: every typed accessor the doc uses (`onConnectionStateChanged`, `onMessageReceived/Sent/Propagated/Error`, `onChannelMessageReceived/Sent/Error`, `createNode`, `start`, `subscribe`, `send`, `channelCreate/Exists/Send`) matches the generated `DeliveryModule` API, and `LogosResult` resolves from `logos_sdk.h` alone (the single include the doc shows is sufficient). Full `nix build` succeeds.
- **Step 8 runtime**: `nix run . -- --user-dir …` launches, `onContextReady` runs, and `createNode({"mode":"Core","preset":"logos.test"})` → `start()` bring up the node successfully (delivery node logs `createNode callback … ret: 0`, `Node started successfully`, libp2p listening, dialing the six `logos.test` bootstrap nodes). The `--user-dir` flag is accepted.

**One correction (inline):** Step 5, line 219 — the `messageSent` → `messagePropagated` order is reversed (it should be `messagePropagated` → `messageSent`, per the pinned header, the v0.2.0 README, and this PR’s own Step 9).

**Not exercised (environment, not a doc issue):** the two-instance message/channel exchange (Steps 9–10). My node never reached a `connected` state — it dialed all six `logos.test` peers but rendezvous kept failing (`could not get a peer supporting WakuRendezVousCodec`) for ~1h, i.e. p2p egress to the testnet isn’t available from my sandbox — so the `connectionStateChanged=="connected"` path that would trigger `subscribe`/`send`/`channel*` never fired. Nothing in the doc to change here.

**Non-blocking upstream FYI (no doc change):** the doc types every event `timestamp` as `qint64`, which is the right intent and compiles — but the module-builder `0.2.5` generated wrapper (`delivery_module_api.cpp`) marshals each one with `_args.at(N).toInt()` into an `int` callback parameter, so on `0.2.5` the ns-since-epoch value is truncated to 32 bits before the lambda runs, regardless of the declared type. Worth an upstream codegen issue against `logos-module-builder`; the doc’s `qint64` is the correct forward-looking type.

PR #437, opened 2026-08-10
fix(blend): /mantle/sdp/declarations returns an object, and clarify the active field

xAlisher:
While following **Join the Blend network as a core node** against a live 0.2.1 testnet node, the `/mantle/sdp/declarations` example did not match the actual response.

## 1. Response is an object, not an array
The doc shows a JSON array `[ { ... } ]`. The endpoint actually returns a JSON **object keyed by declaration id**:

```json
{
  "dd245596...": { "service_type": "BN", "provider_id": "...", "created": 0, "active": 11, ... },
  "427f6463...": { "service_type": "BN", "provider_id": "...", "created": 5, "active": 11, ... }
}
```

(11 entries on the node I checked; each value carries exactly the fields the doc already lists.) The fix updates the example to the object shape and tells the reader to find their entry by `provider_id`/`zk_id`.

## 2. `active` is not a fixed `created + 2`
The doc states `active == created + 2`. That is only true right after activation. `active` is the **latest epoch the node has re-attested activity for** via the periodic *Active* message, so it advances over time. On the live node, declarations created at epochs **0, 4, 5** all show `active = 11` (the current epoch) — not `created + 2`. The fix keeps the ~2-epoch activation delay but corrects the field description.

Both points verified against a running `blockchain_module` 0.2.1 node (same class of check as the earlier `/cryptarchia/info` schema fix). Opened as a draft — happy to adjust wording. Note: I don't have label permissions on this repo, so no labels applied.


weboko (commented):
Dogfooded this against the doc end to end:

- Installed deps and ran `npm run build` (Docusaurus, `onBrokenLinks: 'throw'`) on this branch — **build passes**, so the `[Blend node]` link and the rest still resolve.
- Checked the corrected example: it is well-formed and correctly 2-space nested under `"<DECLARATION_ID>"`, and the `created: 1` / `active: 3` values are internally consistent with the new "equals `created + 2` right after activation" wording.
- Confirmed the removed bullet lost no information — the `zk_id` / `provider_id` → key mappings are carried into the new "find your entry by…" bullet.

Both corrections (object-keyed-by-id, and `active` as a re-attested epoch that advances) match the live-node evidence in the PR description and are self-consistent. Nice catch. One follow-up:

**The same example, with both of the same bugs, still lives in another page** — `docs/run-a-node/get-started/run-logos-node-blockchain-storage-delivery.md` (not in this PR's diff), lines ~329–349:

- lines 329–343 show the array shape `[ { … } ]`
- line 349 says *"`active` is 2 epochs in the future, when the declaration will become active"* — the same `created + 2` misconception this PR corrects.

After this merges the two pages will contradict each other. Suggest applying the identical fix there. Concretely, that block/bullets should become (note the 3-space indent on that page):

```
   # > {
   # >   "<DECLARATION_ID>": {
   # >     "service_type": "BN",
   # >     "provider_id": "35d60d973560b8344f83dc266a3fe89e35a3dcf9959c492d0a7a0b7a85c5d2ce",
   # >     "locked_note_id": "<BLEND_ZK_NOTE_ID>",
   # >     "locators": [
   # >       "/ip4/<YOUR_IP>/udp/<YOUR_BLEND_PORT>/quic-v1"
   # >     ],
   # >     "zk_id": "13cccf99f90fd78c2134891ce3c1afce0605753a7694b9d56678d63a8d471820",
   # >     "created": 1,
   # >     "active": 3,
   # >     "withdraw_at": null,
   # >     "nonce": 0
   # >   }
   # > }
```

and the bullets:

```
   - The response is a JSON **object keyed by declaration id** (not a list). Find your entry by its `provider_id` (your `BlendSigning` key) or `zk_id` (your `BlendZk` key).
   - `service_type: BN` identifies it as a [Blend node](../../get-started/glossary.md#blend-node) declaration.
   - `created` is the epoch your declaration was included; it takes effect about two epochs later. `active` is the most recent epoch your node has re-attested activity for (via the periodic Active message), so it **advances over time** — it equals `created + 2` right after activation and grows on a long-running node.
```

Happy for that to be a separate PR if you'd rather keep this one scoped to the core-node page.

ntn-x2 (approved):
Nice catch, thanks!

xAlisher:
Thanks @weboko — and good catch on the second page. Pushed `087759e` applying the identical correction to `docs/run-a-node/get-started/run-logos-node-blockchain-storage-delivery.md` (array → object-keyed-by-id, and the `active` bullet that said "2 epochs in the future"), so the two pages stay consistent after merge. Also appreciate the `npm run build` / self-consistency pass.

weboko (approved):
Re-dogfooded after the follow-up commit (087759e4), which applies the identical `/mantle/sdp/declarations` correction to the storage/delivery guide that my earlier review asked for. Verified end to end on this branch:

- **Build passes.** `npm run build` (Docusaurus, `onBrokenLinks: 'throw'`) succeeds, so the `[Blend node]` and `[epochs]` links in the edited block still resolve.
- **Example is well-formed.** The corrected block now parses as a JSON **object keyed by `<DECLARATION_ID>`** (not a list), with the fields correctly nested — I round-tripped it through a JSON parser. `created: 1` / `active: 3` is internally consistent with the new "equals `created + 2` right after activation" wording.
- **No contradiction between the two pages.** Both `join-the-blend-network-as-a-core-node.md` and `run-logos-node-blockchain-storage-delivery.md` now show the same object shape, the same `created:1`/`active:3` values, and the same "`active` advances as the node re-attests" explanation. The post-merge inconsistency I flagged is resolved.

Both corrections match the live-node behaviour described in the PR and are self-consistent across the repo now. LGTM.

weboko:
human lgtm too

PR #444, opened 2026-08-11
alisher blend fixes

kashepavadan:


kashepavadan:
@xAlisher just btw, made some clarity/accuracy fixes to your blend additions

PR #450, opened 2026-08-13
v0.2.2 update for Zone SDK

kashepavadan:


docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -67,32 +67,58 @@ A channel is created automatically the first time an operation references a prev
    
 1. Initialize a `ZoneSequencer` and publish the first [inscription](../../get-started/glossary.md#inscription) inside your event loop once `Event::Ready` has fired. The channel is created on-chain automatically, naming this sequencer's key as the sole accredited key.
 
-   ```rust
-   use lb_zone_sdk::{
-       CommonHttpClient, adapter::NodeHttpClient, sequencer::ZoneSequencer,
-   };
+    ```rust
+    use lb_zone_sdk::{
+        CommonHttpClient, adapter::NodeHttpClient, sequencer::{FundingConfig, ZoneSequencer},
```
weboko: The corrected `init_with_config` flow below needs `SequencerConfig` in scope — add it to this import:

```suggestion
        CommonHttpClient, adapter::NodeHttpClient, sequencer::{FundingConfig, SequencerConfig, ZoneSequencer},
```

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -67,32 +67,58 @@ A channel is created automatically the first time an operation references a prev
...
+        // `funding_pk` is the public key of a wallet key that node controls —
+        // the same key the node's own configuration declares as its funding wallet.
+        funding_pk,
+        max_tx_fee: 1_000_000.into(),
+        priority_fee: FundingConfig::DEFAULT_PRIORITY_FEE,
+    };
 
-   `publish` returns synchronously after enqueueing the transaction into the sequencer's pending set; the post reaches the node the next time the event loop polls `next_event`. Persist the returned `PublishResult` and `SequencerCheckpoint` into your outbox.
+    // Initialize the sequencer for this channel
+    let mut sequencer = ZoneSequencer::init(channel_id, signing_key, node, funding, None);
```
weboko: **This won't compile.** In the Zone SDK `0.2.2`, `ZoneSequencer::init` takes four args — `(channel_id, signing_key, node, checkpoint)` — with no `funding` parameter. Funding is set on `SequencerConfig.funding` (`Option<FundingConfig>`) and applied via `init_with_config`. It's also optional: `None` builds fee-less transactions (valid while gas prices are zero), so the "mandatory" wording is inaccurate. This mirrors the reference `tui-zone` impl (`cli_sequencer_config` builds `SequencerConfig { funding: …, ..SequencerConfig::default() }` and calls `init_with_config`).

```suggestion
    // Funding is optional. With a `FundingConfig` set, every publish-type call
    // funds its transaction from the connected node's wallet before signing —
    // the node builds and proves the fee transfer, and the secret key never
    // leaves it. Leaving funding unset (`None`) builds fee-less transactions,
    // which are only valid while gas prices are zero.
    let funding = FundingConfig {

        // `funding_pk` is the public key of a wallet key the node controls —
        // the same key the node's own configuration declares as its funding wallet.
        funding_pk,
        max_tx_fee: 1_000_000.into(),
        priority_fee: FundingConfig::DEFAULT_PRIORITY_FEE,
    };

    // `init` takes no funding argument — funding lives on `SequencerConfig`,
    // so pass a config through `init_with_config`.
    let config = SequencerConfig {
        funding: Some(funding),
        ..SequencerConfig::default()
    };
    let mut sequencer =
        ZoneSequencer::init_with_config(channel_id, signing_key, node, config, None);
```

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -67,32 +67,58 @@ A channel is created automatically the first time an operation references a prev
...
 
-1. (Optional) Reconfigure the channel by calling `sequencer.handle().channel_config(..)` with a [`ChannelConfig`](https://app.notion.com/p/nomos-tech/1-5-0-Mantle-33d261aa09df8051b0d0cd4d5ddade85?source=copy_link#f96261aa09df826a93d801db1e432a54) operation.
+    // Inside the event loop, once `Event::Ready` has fired:
+    // Publishing the first inscription creates the channel just-in-time.
+    let (result, checkpoint) = sequencer.handle().publish(genesis_zone_block)?;
```
weboko: `SequencerHandle::publish` is an `async fn` in `0.2.2` (the reference `tui-zone` calls `sequencer.handle().publish(inscription).await`), so `?` alone won't compile — it needs `.await`:

```suggestion
    let (result, checkpoint) = sequencer.handle().publish(genesis_zone_block).await?;
```

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -67,32 +67,58 @@ A channel is created automatically the first time an operation references a prev
...
+    ```
+
+    `publish` returns synchronously after enqueueing the tx into the sequencer's pending set; the post hits the node the next time the drive loop polls `next_event`. The returned `PublishReceipt` carries everything you need to persist this publish into your outbox alongside the resulting checkpoint.
+
+    :::info
+    To override other sequencer settings, build the config explicitly —
+    `SequencerConfig::new(funding)` fills in the defaults for everything else:
```
weboko: `SequencerConfig::new(...)` doesn't exist — `SequencerConfig` only derives `Default`. Reword to reference `default()`:

```suggestion
    `SequencerConfig::default()` fills in the defaults for everything else:
```

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -67,32 +67,58 @@ A channel is created automatically the first time an operation references a prev
...
+    ```rust
+    use lb_zone_sdk::sequencer::SequencerConfig;
+    let config = SequencerConfig {
+        resubmit_interval: Duration::from_secs(10),
+        ..SequencerConfig::new(funding)
+    };
```
weboko: Two fixes here: `SequencerConfig` is now imported at the top (drop this local `use`), and `SequencerConfig::new(funding)` doesn't exist — set `funding` explicitly and fill the rest with `..SequencerConfig::default()`. This snippet also uses `Duration`, which needs importing:

```suggestion
    use std::time::Duration;
    let config = SequencerConfig {
        funding: Some(funding),
        resubmit_interval: Duration::from_secs(10),
        ..SequencerConfig::default()
    };
```

weboko (commented):
Dogfooded this against the real Zone SDK at tag `0.2.2`: cloned `logos-blockchain/logos-blockchain`, then cross-checked the changed Rust snippets against `zone-sdk/src` (the actual `ZoneSequencer::init` / `init_with_config` / `SequencerConfig` / `FundingConfig` definitions), the SDK's own `zone-sdk/BRIDGING.md`, and the reference `deployment/tui-zone` implementation these tutorials are based on.

**Working well**
- The `OrphanedTx` → `ChannelUpdateTx` rename and its `Inscription` / `AtomicWithdraw` / `Custom(SignedMantleTx)` variants match the crate exactly.
- The new `channel_inscriptions` note matches `DECENTRALIZED_SEQUENCING.md`.
- Every updated spec link (`lip.logos.co/.../bedrock-v1.1-mantle-specification.html#…`) resolves `200`, and every anchor (`message-ordering`, `payload-1`, `channel_deposit`, `channel_withdraw`) exists.
- The round-robin doc's changes are all correct — nothing requested there.

**Blocking — the new funding example in `bridge-assets-between-blockchain-and-zone.md` will not compile as written:**
1. `ZoneSequencer::init` has **no** `funding` parameter (still 4 args: `channel_id, signing_key, node, checkpoint`). Funding lives on `SequencerConfig.funding` (an `Option<FundingConfig>`) and is applied via `init_with_config`.
2. `SequencerConfig::new(funding)` doesn't exist — the type only derives `Default`. Use `SequencerConfig { funding: Some(funding), ..SequencerConfig::default() }`.
3. Funding isn't mandatory — `funding: None` is valid (fee-less txs while gas prices are zero); the doc currently says it's mandatory.
4. `SequencerHandle::publish` is `async` in v0.2.2, so it needs `.await`.

Inline suggestions for each below, modelled on the reference `tui-zone` code (`cli_sequencer_config` → `init_with_config` → `.publish(...).await`). For corroboration, the sibling `inscribe-data-on-chain-using-zone-sdk.md` already uses the correct 4-arg `init`, which is another sign the 5-arg form here is the anomaly.

kashepavadan:
Take a look now @weboko

docs/blockchain/zone-sdk/bridge-assets-between-blockchain-and-zone.md
```diff
@@ -67,32 +67,32 @@ A channel is created automatically the first time an operation references a prev
...
+    // Initialize the sequencer for this channel
+    let mut sequencer = ZoneSequencer::init(channel_id, signing_key, node, None);
 
-   // Inside the event loop, once `Event::Ready` has fired:
-   // Publishing the first inscription creates the channel just-in-time.
-   let (result, checkpoint) = sequencer.handle().publish(genesis_zone_block)?;
-   ```
+    // Inside the event loop, once `Event::Ready` has fired:
+    // Publishing the first inscription creates the channel just-in-time.
+    let (result, checkpoint) = sequencer.handle().publish(genesis_zone_block)?;
```
weboko: Non-blocking nit: `SequencerHandle::publish` is an `async fn` in `0.2.2`, and the sibling `publish_atomic_withdraw` call later in this same doc (the single-sequencer withdrawal step) already ends in `.await?` — so this line is internally inconsistent and, taken literally, won't compile without `.await`:

```suggestion
    let (result, checkpoint) = sequencer.handle().publish(genesis_zone_block).await?;
```

FYI the same omission exists upstream in `zone-sdk/BRIDGING.md` (which uses `.await` on `publish_atomic_withdraw` but not on `publish`), so it's worth fixing at the source too. Not a blocker for this PR.

weboko (commented):
Re-dogfooded after the `match to v0.2.2` commit (head `b72c7a7`), against the Zone SDK at tag `0.2.2`.

**All previously-blocking issues are resolved:**
- The invalid funding example is gone — the main block is back to the correct 4-arg `ZoneSequencer::init(channel_id, signing_key, node, None)`, matching the SDK's own `zone-sdk/BRIDGING.md` exactly.
- The `SequencerConfig::new(...)` / `init_with_config` info box (which referenced APIs that don't exist in that form) has been removed.
- The `PublishReceipt` narrative is now accurate.

**Re-verified the rest against the crate source:**
- `OrphanedTx` → `ChannelUpdateTx` and the `Inscription` / `AtomicWithdraw` / `Custom(SignedMantleTx)` variants — correct.
- `PublishResult.tx` → `PendingTx::AtomicWithdraw` / `PendingTx::Inscription` (withdrawal steps) — correct (`PublishResult { tx: PendingTx }`, variants match).
- The `channel_inscriptions` note in the round-robin doc — correct.
- Every updated `lip.logos.co/.../bedrock-v1.1-mantle-specification.html#…` link resolves `200` with a valid anchor.

Looks good to me. One remaining minor, non-blocking inconsistency inline — fine to fix now or leave to match upstream.

weboko (approved):
manual review lgtm!

PR #452, opened 2026-08-14
chore: revise storage tutorials

gmega:
This PR revises the storage tutorials so that:

1. headless tutorials use lgpd instead of nix builds;
2. the storage node tutorial include the download of public files available on the testnet;
3. NAT configuration reflects our more modern NAT traversal machinery;
4. the UI tutorial removes the Mix configs (those are baked into the UI binary as of https://github.com/logos-co/logos-storage-ui/pull/84).

In particular, I've strived to keep configurations as compact as possible, and avoid useless steps. I kept the nix build and the standalone app in the UI tutorial for now, as it's provided in addition to the simpler, basecamp-based option.

### DIFF
--- a/docs/storage/get-started/run-logos-storage-node.md
+++ b/docs/storage/get-started/run-logos-storage-node.md
@@ -215,7 +215,7 @@ We will now download the Logos book, [Farewell to Westphalia](https://logos.co/b
     ```sh
     CID="zDvZRwzkzrrYB6sS1rRpRLt4gBhc1pWoyTSjkfszfmj1seaYYLCZ"
     logoscore call storage_module downloadToUrl "$CID" "$(pwd)/farewell-to-westphalia.pdf" false 65536
-    ``
+    ```
 
     :::tip
     The `local` flag reads only from locally cached data when set to `true`; `false` fetches from the network.
@@ -224,7 +224,7 @@ We will now download the Logos book, [Farewell to Westphalia](https://logos.co/b
 1.  Wait for a while for the file to download. After a few seconds, check if the downloaded file is present at the destination path. You should try to open the pdf, and it should contain the whole book.
 
     ```sh
-    sha1sum "$(pwd)/farewell-to-westphalia.pdf"
+    shasum "$(pwd)/farewell-to-westphalia.pdf"
     # 2c6b4dc8e8e4dae336b87b9922c38f3c94217872  farewell-to-westphalia.pdf
     ```
 
--- a/docs/storage/get-started/run-logos-storage-node.md
+++ b/docs/storage/get-started/run-logos-storage-node.md
@@ -214,7 +214,7 @@ We will now download the Logos book, [Farewell to Westphalia](https://logos.co/b
 
     ```sh
     CID="zDvZRwzkzrrYB6sS1rRpRLt4gBhc1pWoyTSjkfszfmj1seaYYLCZ"
-    logoscore call storage_module downloadToUrl "$CID" "$(pwd)/farewell-to-westphalia.pdf" true 65536
+    logoscore call storage_module downloadToUrl "$CID" "$(pwd)/farewell-to-westphalia.pdf" false 65536
     ``
 
     :::tip
--- a/docs/storage/get-started/run-logos-storage-node.md
+++ b/docs/storage/get-started/run-logos-storage-node.md
@@ -23,17 +23,23 @@ This procedure covers how to build and run the [Logos Storage Module](https://gi
 
 - A supported OS:
     - Linux
+    - Mac OS (should work, but not tested)
 - `jq` on your `PATH`.
     - To verify, run: `jq --version`
-- **Nix** with flakes enabled. Install from [nixos.org](https://nixos.org/download.html), then enable flakes:
+- The Logos tool suite:
+    - [`logoscore`](https://github.com/logos-co/logos-logoscore-cli/releases/tag/0.2.2) (the Logos runtime);
+    - [`lgpd`](https://github.com/logos-co/logos-package-downloader/releases/tag/0.2.1) (the Logos package downloader);
+    - [`lgpm`](https://github.com/logos-co/logos-package-manager/releases/tag/0.2.1) (the Logos package manager).
 
-   ```bash
-   mkdir -p ~/.config/nix
-   echo 'experimental-features = nix-command flakes' >> ~/.config/nix/nix.conf
-   ```
-- [`logoscore`](https://github.com/logos-co/logos-logoscore-cli/releases/tag/0.2.2), and [`lgpm`](https://github.com/logos-co/logos-package-manager/releases/tag/0.2.1) installed. To install these tools, use the `install-node-tools.sh` helper script:
+  You can obtain them by running:
 
     ```bash
+    # Export those first or the script will fetch the latest version, which might not
+    # work with this tutorial
+    export LGPM_TAG=0.2.1
+    export LGPD_TAG=0.2.1
+    export LOGOSCORE_TAG=0.2.2
+
     curl -fsSL https://raw.githubusercontent.com/logos-co/logos-docs/main/resources/scripts/install-node-tools.sh | sh
     export PATH="$PWD/bin:$PATH"
     ```
@@ -41,41 +47,44 @@ This procedure covers how to build and run the [Logos Storage Module](https://gi
 
 ## What to expect
 
-- You can connect a [Logos Storage](../../get-started/glossary.md#logos-storage) node to the testnet and have it listed among the bootstrap peers.
-- You can publish a file to the network and retrieve a content address to share with other nodes.
-- You can download the file back from the network and confirm it lands on disk.
+In this tutorial, you will:
 
-## Build and install the storage module
+- Connect a [Logos Storage](../../get-started/glossary.md#logos-storage) node to the testnet.
+- Publish a file to the network.
+- Download an existing file - the Logos book, [Farewell to Westphalia](https://logos.co/book) - from the Logos storage network.
 
-1.  Build the module package with Nix:
+## Download and install the storage module
+
+1.  Download the storage module:
 
     ```sh
-    nix build 'github:logos-co/logos-storage-module/v2.1.0#lgx-portable' -o storage-lgx
+    mkdir -p storage-lgx
+    lgpd download storage_module --version 2.1.2 -o storage-lgx
     ```
 
-    - This produces a `.lgx` package in `./storage-lgx/`.
+    This should download an `lgx` file in the `storage-lgx` folder.
 
-    :::info
-    The initial Nix build takes 15–20 minutes on first run. Subsequent builds use the Nix cache and complete in seconds.
-    :::
-
-1.  Install the package into a local modules directory using `lgpm`. The package is a local build and is unsigned, so pass `--allow-unsigned`:
+1.  Install the package using `lgpm`.
 
     ```sh
     mkdir -p modules
-    lgpm --modules-dir ./modules --allow-unsigned install --file storage-lgx/*.lgx
+    lgpm --modules-dir ./modules install --file storage-lgx/*.lgx
     ```
 
 1.  Confirm the module landed:
 
     ```sh
     lgpm --modules-dir ./modules list
-    # storage_module appears in the listing
+    Found 1 installed module(s):
+
+    NAME                           VERSION         TYPE       CATEGORY
+    ----------------------------------------------------------------------
+    storage_module                 (v)           core       protocol
     ```
 
 ## Start the daemon and load the storage module
 
-Run `logoscore` with the modules directory, then load and initialise the storage module against the testnet config.
+Run `logoscore` with the modules directory, then load and initialise the storage module.
 
 Several module calls in this procedure are **asynchronous**: the call returns `"result":true` as soon as the command is accepted, and the real outcome is delivered later as an event (`storageStart`, `storageUploadDone`, `storageDownloadDone`, `storageRemoveDone`, `storageDownloadManifestDone`). These events are emitted to event subscribers (such as the Storage UI); the `logoscore call` client does not subscribe to them, so they do **not** appear in `logs.txt`. Each step below instead waits briefly and confirms the outcome with a follow-up query (for example `manifests` or `exists`).
 
@@ -122,43 +131,32 @@ Several module calls in this procedure are **asynchronous**: the call returns `"
 
     - To see every method the module exposes (the same methods you can `call`), run `logoscore module-info storage_module`.
 
-1.  Create the storage config. Use **absolute** paths: in daemon mode the module runs as its own process, whose working directory is not the one you are typing in, so relative paths resolve to the wrong place. Replace `your-public-IP` by your public external IP. The `$(pwd)` in the heredoc takes care of it:
+1.  Create a minimal storage config. Use **absolute** paths: in daemon mode the module runs as its own process, whose working directory is not the one you are typing in, so relative paths resolve to the wrong place. The `$(pwd)` in the heredoc takes care of it:
 
     ```sh
     mkdir -p "$(pwd)/storage-data"
     cat > config.json <<EOF
     {
         "data-dir": "$(pwd)/storage-data",
-        "log-level": "INFO",
-        "log-file": "$(pwd)/storage-data/storage.log",
-        "listen-ip": "0.0.0.0",
-        "listen-port": 8091,
-        "disc-port": 8090,
-        "network": "logos.test",
-        "nat": "extip:<your-public-IP>"
+        "log-file": "$(pwd)/storage-data/storage.log"
     }
     EOF
     ```
-
-    - With `"nat": "extip:<your-public-IP>"` the node announces the machine's own IP as-is. If it is reachable from other peers over Internet (with ports forwarded for example), the node will be able to upload and download from other peers. Otherwise, it can only download from the network but other peers will not be able to download from this node. To know more about node reachable state: see [Connectivity](../concepts/connectivity.md).
 
     - `config.json` includes the following fields:
 
     | Field | Purpose |
     |-------|---------|
     | `data-dir` | Storage repository path (absolute) |
-    | `log-level` | Log verbosity |
     | `log-file` | Node log destination (absolute) |
-    | `listen-ip` | Local TCP bind address |
-    | `listen-port` | Public TCP libp2p port |
-    | `disc-port` | Public UDP discovery port |
-    | `network` | Storage network preset |
-    | `nat` | Public IP advertisement mode — see [Connectivity](../concepts/connectivity.md) |
 
-    - Every omitted key keeps its default: the node joins the `logos.test` network preset (which provides the testnet bootstrap settings), binds discovery to the default UDP port `8090`, and picks a random TCP `listen-port`.
-    - For a public, reachable node, set fixed `listen-port` (TCP) and `disc-port` (UDP) values: see [Connectivity](../concepts/connectivity.md).
+    - The default settings for Logos storage should be enough to get your node properly connected onto the Logos testnet. In case you want more control over port allocation, or want to learn more about how Logos storage operates, see [Connectivity](../concepts/connectivity.md).
 
-1.  Initialise the storage module with the testnet configuration. `init` is synchronous and returns `true` on success (the `@config.json` syntax loads the file's contents as the argument):
+    :::tip
+    If you plan on running a node for longer, consider helping the network by setting up port mapping on your router (see [Connectivity](../concepts/connectivity.md) for details).
+    :::
+
+1.  Initialise the storage module. `init` is synchronous and returns `true` on success (the `@config.json` syntax loads the file's contents as the argument):
 
     ```sh
     logoscore call storage_module init @config.json
@@ -171,15 +169,18 @@ Several module calls in this procedure are **asynchronous**: the call returns `"
     # Wait few seconds to start
     ```
 
-1.  Inspect the running node with `debug`. It returns the node's identity: its `id` ([peer ID](../../get-started/glossary.md#peer-id)) and its `spr`, the signed record other nodes use to connect to you (see [Connectivity](../concepts/connectivity.md)):
+1.  Inspect the running node with `debug`. It returns a lot of information about the node, including its `id` ([peer ID](../../get-started/glossary.md#peer-id)) and its `spr`, the signed record other nodes use to connect to you (see [Connectivity](../concepts/connectivity.md)):
 
     ```sh
-    logoscore call storage_module debug
+    logoscore call storage_module debug | jq .result.value.id
+    # "16Uiu2HAmMA4NuQoCHz9p7jUskVjDd8WncwG3p6qBNnhnftUE5Q9C" # Your peer ID
+    logoscore call storage_module debug | jq .result.value.spr
+    # "spr:CiUIAhIhA35P5KZosVyfWTfIHBVtC_PtI ... H9gX-vA" # Your SPR
     ```
 
-## Publish and download a file
+## Publish a file
 
-Once the node is running and connected to the testnet, publish a file and verify the round-trip.
+We will now publish a file to the Logos storage network. We create a simple file just for this tutorial, but you could publish any file you'd like.
 
 1.  Create a file to publish:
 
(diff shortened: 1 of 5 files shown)


docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -205,22 +206,26 @@ Once the node is running and connected to the testnet, publish a file and verify
...
+We will now download the Logos book, [Farewell to Westphalia](https://logos.co/book), from the Logos storage network.
+
+1. We will use `downloadToUrl` to download the file from the network and place it into your local disk. It takes the CID, an **absolute** destination path, a `local` flag, and a chunk size in bytes. We set `local` to `false` as this file is not currently available in your node. Like `uploadUrl` it runs in the background and completes with a `storageDownloadDone` event:
 
     ```sh
-    logoscore call storage_module downloadToUrl "$(cat cid.txt)" "$(pwd)/hello-destination.txt" true 65536
-    ```
+    CID="zDvZRwzkzrrYB6sS1rRpRLt4gBhc1pWoyTSjkfszfmj1seaYYLCZ"
+    logoscore call storage_module downloadToUrl "$CID" "$(pwd)/farewell-to-westphalia.pdf" true 65536
+    ``
```
weboko: **Two bugs on these two lines: the `local` flag is inverted, and the fence is closed with two backticks instead of three.**

**1. `local` must be `false` here.** The prose directly above says *"We set `local` to `false` as this file is not currently available in your node"*, and the tip below says `true` reads only local data — but the command passes `true`. The module's own API doc agrees:

```
downloadToUrl(cid, filePath, local, chunkSize)
  `local` - if true, only reads from locally cached data (no network).
```

Ran it exactly as written against storage_module 2.1.2 — it fails immediately, no file created:

```
{"method":"downloadToUrl","result":{"error":"Failed to start download.","success":false,"value":null}}
```

The node log confirms the cause is that it never goes to the network:
```
ERR Failed to MANIFEST. error="Failed to fetch manifest zDvZRwzk... : Key doesn't exist"
```

As a control, `local=true` against a CID I *had* uploaded locally succeeded and round-tripped the file correctly — so the flag behaves exactly as documented, and `true` is simply the wrong value for a file you don't have yet.

**2. Line 218 breaks the rest of the page.** `` `` `` doesn't close the ```` ```sh ```` opened on line 215, so fence parity inverts for everything below. I built the site (`yarn build`) — it passes with no error, and the rendered page silently shows the `:::tip` admonition as *literal text inside a code block*, swallowing the step-1 prose and the `sha1sum` block with it:

```
logoscore call storage_module downloadToUrl "$CID" ... true 65536
``
:::tip
The `local` flag reads only from locally cached data when set to `true`; ...
:::
Wait for a while for the file to download. ...
sha1sum "$(pwd)/farewell-to-westphalia.pdf"
```

The knock-on leaves an unclosed fence at line 274. Both fixed together:

````suggestion
    logoscore call storage_module downloadToUrl "$CID" "$(pwd)/farewell-to-westphalia.pdf" false 65536
    ```
````

One thing I could **not** verify: whether the download then actually succeeds. My sandbox can't reach the testnet — every peer in the DHT table stays `seen=false` and discv5 `GetProviders` times out throughout — so I can't confirm the CID resolves or that the `sha1sum` on line 228 matches. Worth someone re-running that on a connected machine.

File: docs/storage/get-started/run-logos-storage-node.md [hunk 2]
```diff
@@ -23,59 +23,68 @@ This procedure covers how to build and run the [Logos Storage Module](https://gi
...
-   ```
-- [`logoscore`](https://github.com/logos-co/logos-logoscore-cli/releases/tag/0.2.2), and [`lgpm`](https://github.com/logos-co/logos-package-manager/releases/tag/0.2.1) installed. To install these tools, use the `install-node-tools.sh` helper script:
+  You can obtain them by running:
 
     ```bash
+    # Export those first or the script will fetch the latest version, which might not
+    # work with this tutorial
+    export LGPM_TAG=0.2.1
+    export LGPD_TAG=0.2.1
+    export LOGOSCORE_TAG=0.2.2
```
weboko, line 41: These exports have no effect — the comment's justification is not accurate. `install-node-tools.sh` assigns the tags **unconditionally** (`resources/scripts/install-node-tools.sh:10-12`):

```sh
LOGOSCORE_TAG=0.2.2
LGPD_TAG=0.2.1
LGPM_TAG=0.2.1
```

They're plain assignments, not `${VAR:-default}`, so an exported `LGPM_TAG` is overwritten before use. The script never "fetches the latest version" — it's already pinned, and it pins to exactly the versions listed here. I confirmed by running the block as written: the installed tools are `logoscore 0.2.2`, `lgpd 0.2.1`, `lgpm 0.2.1`, which is what you'd get with no exports at all.

Two ways to resolve, both fine:

- **Drop the block** (simplest, since the script is already pinned):
  ```suggestion
    curl -fsSL https://raw.githubusercontent.com/logos-co/logos-docs/main/resources/scripts/install-node-tools.sh | sh
  ```
  (then delete the now-orphaned blank line above the `curl`)

- **Or make the script honour the env vars**, which would make this block do what it claims and give readers a real pin against future script bumps:
  ```sh
  LOGOSCORE_TAG=${LOGOSCORE_TAG:-0.2.2}
  LGPD_TAG=${LGPD_TAG:-0.2.1}
  LGPM_TAG=${LGPM_TAG:-0.2.1}
  ```
  That's a change to `resources/scripts/install-node-tools.sh`, outside this PR's diff — flagging rather than suggesting inline.

I'd lean toward the second, since the script is fetched from `main` at run time and will drift from this tutorial the moment someone bumps it.

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -23,59 +23,68 @@ This procedure covers how to build and run the [Logos Storage Module](https://gi
...
 1.  Confirm the module landed:
 
     ```sh
     lgpm --modules-dir ./modules list
-    # storage_module appears in the listing
+    Found 1 installed module(s):
+
+    NAME                           VERSION         TYPE       CATEGORY
+    ----------------------------------------------------------------------
+    storage_module                 (v)           core       protocol
```
weboko, line 82: The version column is garbled — `(v)` isn't what the tool prints. Real output from `lgpm 0.2.1` after the install step above:

```
NAME                           VERSION         TYPE       CATEGORY       
----------------------------------------------------------------------
storage_module                 2.1.2           core       protocol
```

Looks like a version placeholder that didn't get substituted. Since the download step right above pins `--version 2.1.2`, the concrete value is both correct and useful here:

```suggestion
    storage_module                 2.1.2           core       protocol
```

Same line needs the same fix in `run-mix-network-of-storage-nodes.md:80`.

Related, and just outside this diff: the `logoscore status` samples on lines 111-112 and 128-129 have the same artifact fused onto the module name (`storage_modulev`, `capability_modulev`). Real output is:
```
  storage_module     v2.1.2  loaded      21s
  capability_module  v1.0.0  loaded      29s
```
Worth fixing in the same pass while you're in here.

docs/storage/get-started/run-mix-network-of-storage-nodes.md
```diff
@@ -24,55 +24,60 @@ This procedure stands up a small local [Mix](../concepts/mix.md) network using `
...
 ## What to expect
 
 - You can run several `logoscore` daemons side by side with separate `--config-dir`s.
 - You can set up a private Mix network and configure storage nodes to anonymize their lookups through it.
 - You can exchange a file between two storage nodes and verify the content lookup was tunnelled over Mix.
 
-## Build and install the storage module
-
-1.  Build the module package with Nix:
+1.  Download the storage module:
```
weboko, line 58: **The `## Build and install the storage module` heading was dropped, so these three steps are now orphaned under `## What to expect`.**

That section (line 50) is a bulleted overview of outcomes, and the module download/install/verify steps now render as part of it — the next heading isn't until `## Launch the bootstrap Mix node (node 1)` on line 83. The equivalent section still exists in `run-logos-storage-node.md`, so the two tutorials have diverged structurally.

The old heading said "Build and install", which no longer fits now that it's an `lgpd` download rather than a Nix build:

```suggestion
## Download and install the storage module

1.  Download the storage module:
```

That also matches the heading used in `run-logos-storage-node.md:56`, keeping the two pages parallel.

I ran these three steps as written and they work — `lgpd download storage_module --version 2.1.2` fetches `storage_module-2.1.2.lgx` and `lgpm install` lands it. Only the heading is missing.

docs/storage/get-started/run-mix-network-of-storage-nodes.md
```diff
@@ -24,55 +24,60 @@ This procedure stands up a small local [Mix](../concepts/mix.md) network using `
...
 1.  Confirm the module landed:
 
     ```sh
     lgpm --modules-dir ./modules list
-    # storage_module appears in the listing
+    Found 1 installed module(s):
+
+    NAME                           VERSION         TYPE       CATEGORY
+    ----------------------------------------------------------------------
+    storage_module                 (v)           core       protocol
```
weboko, line 82: Same `(v)` placeholder as `run-logos-storage-node.md:82` — `lgpm list` prints the real version:

```suggestion
    storage_module                 2.1.2           core       protocol
```

weboko (commented):
Dogfooded the headless tutorial end-to-end on a clean Ubuntu 24.04 container: ran the install script, downloaded and installed storage_module 2.1.2 via `lgpd`/`lgpm`, started the daemon, loaded and initialised the module, uploaded a file, pulled its CID, and ran the full teardown. Also built the site (`yarn build`) to check rendering.

**The lgpd/lgpm switch is a clear improvement** — it replaces a 15-20 minute Nix build with a download that takes seconds, and every step through "Publish a file" worked exactly as written, as did the whole "Remove content and shut everything down" sequence (`remove` → `exists` returns `false` → `stop` → `destroy` → daemon `not_running`). Nice change.

Four things to fix, one of them blocking:

1. **The Farewell to Westphalia download can't work as written** (line 217). It passes `local=true`, which the API defines as "only reads from locally cached data (no network)" — for a file you don't have yet that always fails, and it does: `"error":"Failed to start download."`, no file written. The prose two lines up already says it should be `false`.
2. **Line 218 closes a code fence with two backticks instead of three**, which inverts fence parity for the rest of the page. `yarn build` passes silently, but the rendered page shows the `:::tip` admonition as literal text swallowed inside a code block, along with the step prose and the `sha1sum` block.
3. **`lgpm list` sample output shows `(v)` instead of `2.1.2`** — an unsubstituted placeholder, in both files.
4. **The mix tutorial lost its `## Build and install the storage module` heading**, orphaning three steps under `## What to expect`.

Also worth a look, outside the strict diff: the `export LGPM_TAG=...` block in the prerequisites is a no-op, because `install-node-tools.sh` assigns those tags unconditionally rather than with `:-` defaults. Details inline on line 37.

**What I could not verify:** anything requiring testnet connectivity. My sandbox never reached a peer — all DHT table entries stayed `seen=false` and discv5 `GetProviders` timed out continuously — so I can't confirm the book CID on line 216 resolves, nor the `sha1sum` on line 228. Those two need a re-run on a connected machine. Everything else above was reproduced directly.

For the UI tutorial: the removals look right to me and I have no findings, though I couldn't exercise the GUI here. One thing to sanity-check is that dropping the port-forwarding prerequisite is safe for readers on restrictive NATs now that the AutoNAT warning is gone too — that's your call, and the PR description suggests it's deliberate.

weboko:
left some comments with my ai co-reviewer, lmk if makes sense

docs/storage/get-started/run-logos-storage-node.md
2-towns: We could also create tags for this `logos-docs`. This is something that we need to do ASAP, because it is a general need for all the documentation, and then we can use the Github tag url to point on the bash script for this specific tag version.

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -122,43 +131,32 @@ Several module calls in this procedure are **asynchronous**: the call returns `"
...
-    | `network` | Storage network preset |
-    | `nat` | Public IP advertisement mode — see [Connectivity](../concepts/connectivity.md) |
 
-    - Every omitted key keeps its default: the node joins the `logos.test` network preset (which provides the testnet bootstrap settings), binds discovery to the default UDP port `8090`, and picks a random TCP `listen-port`.
-    - For a public, reachable node, set fixed `listen-port` (TCP) and `disc-port` (UDP) values: see [Connectivity](../concepts/connectivity.md).
+    - The default settings for Logos storage should be enough to get your node properly connected onto the Logos testnet. In case you want more control over port allocation, or want to learn more about how Logos storage operates, see [Connectivity](../concepts/connectivity.md).
 
-1.  Initialise the storage module with the testnet configuration. `init` is synchronous and returns `true` on success (the `@config.json` syntax loads the file's contents as the argument):
+    :::tip
+    If you plan on running a node for longer, consider helping the network by setting up port mapping on your router (see [Connectivity](../concepts/connectivity.md) for details).
```
2-towns, line 156: This is nice.

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -205,22 +206,26 @@ Once the node is running and connected to the testnet, publish a file and verify
        | jq -er '.result.value[0].cid' > cid.txt
     ```
 
-1.  Download the file back from local storage with `downloadToUrl`. It takes the CID, an **absolute** destination path, a `local` flag, and a chunk size in bytes. With `local` set to `true`, the download reads the blocks straight back out of this node's own repository. Like `uploadUrl` it runs in the background and completes with a `storageDownloadDone` event:
+## Download Farewell to Westphalia
+
+We will now download the Logos book, [Farewell to Westphalia](https://logos.co/book), from the Logos storage network.
```
2-towns, line 211: Should we mention that the book was already uploaded into a node and this node is available across the network (Storage node) ?

docs/storage/get-started/run-logos-storage-node.md
2-towns: Agree for the `false` value, but the download is actually verified in the steps below.

docs/storage/get-started/run-logos-storage-node.md
gmega: This is on purpose so we don't have to update this at every release.

docs/storage/get-started/run-mix-network-of-storage-nodes.md
gmega: Should stay that way.

docs/storage/get-started/run-logos-storage-node.md
gmega: Yeah, probably a good idea.

docs/storage/get-started/run-logos-storage-node.md
gmega: Not sure I understand what tags are you thinking about, but we need to be able to pin those versions. I'm gonna make a PR to that install-node-tools so it allows me to override the vars and the above works.

docs/storage/get-started/run-logos-storage-node.md
gmega: Indeed the flag should be false as the text immediately above it states, and we're missing a backtick (but why use two sentences when you can say that same thing in 10 paragraphs??)

docs/storage/get-started/run-mix-network-of-storage-nodes.md
gmega: Ooops, forgot to replace with the correct title.

docs/storage/get-started/run-logos-storage-node.md
gmega: Oh the script is in this repo. I'll just change it as part of this PR.

docs/storage/get-started/run-logos-storage-node.md
gmega: On second thought, I'll deliberately leave this point vague cause I don't want to be saying we're operating all storage nodes. :-D

docs/storage/get-started/run-logos-storage-node.md
gmega: Indeed the flag should be false as the text immediately above it states, and we're missing a backtick (but why use two sentences to describe that when you can do the same thing with 10 paragraphs??)

gmega (commented):
Thanks, I think I've addressed all the comments.

docs/storage/get-started/run-logos-storage-node.md
```diff
@@ -205,22 +206,26 @@ Once the node is running and connected to the testnet, publish a file and verify
...
+
+1. We will use `downloadToUrl` to download the file from the network and place it into your local disk. It takes the CID, an **absolute** destination path, a `local` flag, and a chunk size in bytes. We set `local` to `false` as this file is not currently available in your node. Like `uploadUrl` it runs in the background and completes with a `storageDownloadDone` event:
 
     ```sh
-    logoscore call storage_module downloadToUrl "$(cat cid.txt)" "$(pwd)/hello-destination.txt" true 65536
-    ```
+    CID="zDvZRwzkzrrYB6sS1rRpRLt4gBhc1pWoyTSjkfszfmj1seaYYLCZ"
+    logoscore call storage_module downloadToUrl "$CID" "$(pwd)/farewell-to-westphalia.pdf" false 65536
+    ``
```
weboko: The fence here is still two backticks, so the ```sh block opened on line 215 never closes. This one survived `64a84fd1`, which fixed the `true` -> `false` on the line directly above it.

I confirmed the damage on this branch: `npm run build` **succeeds**, so CI will not flag it, but the built page pulls the tip admonition into the code block as literal text:

```
...false 65536``:::tipThe `local` flag reads only from locally cached data when set to `true`; `false` fetches from the network.:::
```

The resulting fence-parity flip also knocks the `sha1sum` block on lines 226-229 out of its code formatting.

````suggestion
    ```
````

**Review by weboko — COMMENTED** [review]
**Re-review after `64a84fd1` ("address comments"), with a full end-to-end run of both headless tutorials.**

Dogfooded on Linux x86_64 using the pinned tool suite built by this branch's `install-node-tools.sh` (`logoscore` 0.2.2, `lgpd`/`lgpm` 0.2.1, `storage_module` 2.1.2).

### What passed

**`run-mix-network-of-storage-nodes.md` — passed 100% end-to-end, verbatim.** All six nodes came up, `mix-pool.json` and `mix-proxies.json` built with no missing/null fields, the node 5 -> node 6 transfer returned the right content, the SURB grep found the relay selection (4 hits in both `storage-data/node-6/storage.log` and `logs-6.txt`), and the teardown printed exactly `0`. The new `## Download and install the storage module` heading fixes the orphaned-steps problem from the last round.

**`run-logos-storage-node.md` — works verbatim up to and including the publish step:** `lgpd download` -> `lgpm install` -> `lgpm list` -> daemon start -> `load-module` -> `init` -> `start` -> `debug` (peer ID + SPR) -> `uploadUrl` -> `manifests`/`cid.txt`. The teardown section also matches reality exactly, including `exists` printing `false` and `logoscore status` reporting `Status: not_running`.

**`install-node-tools.sh` — the `${VAR:-default}` change is correct in both directions.** Verified with `sh -x`: with no exports it pins `0.2.2`/`0.2.1`/`0.2.1`; with `LGPM_TAG=9.9.9` it correctly rewrites the download URL. That closes the override issue.

### Still open from the last round

The two-backtick fence on line 218 survived this commit — see inline. It's the one blocker: `npm run build` passes, so CI won't catch it, but the rendered page is visibly broken.

### One more, outside the diff hunks (so not inline-commentable)

The `logoscore status` samples on **lines 111-112 and 128-129** have the version placeholder collapsed into the *name* column, so they read as a module called `storage_modulev`:

```
#   storage_modulev         not_loaded  -
#   capability_modulev         loaded      -
```

Real interactive output on 0.2.2:

```
  storage_module     v2.1.2  not_loaded  -
  capability_module  v1.0.0  loaded      15s
```

Unlike the `(v)` in the `lgpm list` VERSION column — which I agree is fine as a deliberate placeholder — this one changes the apparent *module name*. Suggest `storage_module     (v)     not_loaded  -` and `capability_module  v1.0.0  loaded      -`; `capability_module`'s version is independent of storage releases, so it needs no placeholder.

(Also note `logoscore status` only prints this human-readable form on a TTY — it emits JSON when redirected. The samples are right for interactive use, which is how the tutorial reads, so no change needed.)

### What I could not verify here

- **The Farewell to Westphalia download (lines 213-229).** The node behaved correctly and the defaults do ship bootstrap peers exactly as line 153 claims — it dialled all six testnet bootstrap nodes — but my sandbox only allows egress on 443, so every libp2p dial to `tcp/8070`/`udp/9090` failed and `downloadToUrl` returned `Failed to start download.` with 0 connections. **Not a doc problem.** I exercised the same code path locally through the Mix tutorial instead, where `downloadToUrl ... false 65536` succeeded and returned the CID — which independently confirms `false` is the correct flag value for a network fetch.
- **The UI tutorial's Option B (`nix build`).** Nix installed fine and the documented `git clone --recurse-submodules` worked, but the flake's inputs resolve to other `logos-co` repos my session isn't scoped to, so the build 403s. Worth noting the flake pins `storage_module` to `v2.1.2`, consistent with the `--version 2.1.2` in both headless tutorials. Option A needs a GUI, so I could not exercise that either.

File: docs/storage/get-started/run-logos-storage-node.md [hunk 9]
```diff
@@ -23,59 +23,68 @@ This procedure covers how to build and run the [Logos Storage Module](https://gi
...
-   ```bash
-   mkdir -p ~/.config/nix
-   echo 'experimental-features = nix-command flakes' >> ~/.config/nix/nix.conf
-   ```
-- [`logoscore`](https://github.com/logos-co/logos-logoscore-cli/releases/tag/0.2.2), and [`lgpm`](https://github.com/logos-co/logos-package-manager/releases/tag/0.2.1) installed. To install these tools, use the `install-node-tools.sh` helper script:
+  You can obtain them by running:
 
     ```bash
+    # Export those first or the script will fetch the latest version, which might not
+    # work with this tutorial
```
weboko, line 38: Now that `64a84fd1` has made the tags overridable, this comment is the last inaccurate piece: the script **pins** these versions by default — it never "fetches the latest version". Verified on this branch with `sh -x` and no exports set:

```
+ LOGOSCORE_TAG=0.2.2
+ LGPD_TAG=0.2.1
+ LGPM_TAG=0.2.1
```

So for these exact values the exports are a no-op today. Keeping them is still worthwhile as an explicit pin for when the script's defaults move on — just not for the reason stated.

```suggestion
    # The script already pins these versions; set them explicitly so this tutorial
    # keeps working if the script's defaults move on.
```

docs/storage/get-started/run-mix-network-of-storage-nodes.md
```diff
@@ -24,55 +24,62 @@ This procedure stands up a small local [Mix](../concepts/mix.md) network using `
...
-   ```bash
-   mkdir -p ~/.config/nix
-   echo 'experimental-features = nix-command flakes' >> ~/.config/nix/nix.conf
-   ```
-- [`logoscore`](https://github.com/logos-co/logos-logoscore-cli/releases/tag/0.2.2), and [`lgpm`](https://github.com/logos-co/logos-package-manager/releases/tag/0.2.1) installed. To install these tools, use the `install-node-tools.sh` helper script:
+  You can obtain them by running:
 
     ```bash
+    # Export those first or the script will fetch the latest version, which might not
+    # work with this tutorial
```
weboko, line 39: Same inaccuracy as in `run-logos-storage-node.md`: after `64a84fd1` the script pins these versions by default rather than fetching the latest, so the stated reason no longer holds (verified with `sh -x`).

```suggestion
    # The script already pins these versions; set them explicitly so this tutorial
    # keeps working if the script's defaults move on.
```

resources/scripts/install-node-tools.sh
```diff
@@ -7,9 +7,9 @@
...
 # Pinned tool releases.
-LOGOSCORE_TAG=0.2.2
-LGPD_TAG=0.2.1
-LGPM_TAG=0.2.1
+LOGOSCORE_TAG=${LOGOSCORE_TAG:-0.2.2}
+LGPD_TAG=${LGPD_TAG:-0.2.1}
+LGPM_TAG=${LGPM_TAG:-0.2.1}
 
 os=$(uname -s | tr '[:upper:]' '[:lower:]')
 if [ "$os" = darwin ]; then os=macos; fi
```
weboko, line 15: While the tags are being touched here: this maps `darwin` -> `macos`, but `arch` stays `x86_64` on an Intel Mac — and no `*-x86_64-macos.tar.gz` asset exists for any of the three tools. Each release publishes only `aarch64-linux`, `aarch64-macos` and `x86_64-linux`.

Both tutorials list "Mac OS (should work, but not tested)" without qualification, so an Intel Mac user gets a 404 that surfaces as this, with no hint as to why:

```
curl: (22) The requested URL returned error: 404
gzip: stdin: unexpected end of file
tar: Child returned status 1
```

I hit byte-identical output from a transient 502 during this review, so the platform gap is indistinguishable from a network blip. The script header already says "macOS (Apple Silicon)" — making that explicit fails fast instead:

```suggestion
if [ "$os" = darwin ]; then
  os=macos
  if [ "$(uname -m)" != arm64 ]; then
    echo "unsupported platform: macOS on $(uname -m) (only Apple Silicon builds are published)" >&2
    exit 1
  fi
fi
```

PR #166, opened 2026-02-03
feat: initial documentation for Logos Storage

gmega:
This PR adds:

1. a welcome page for Logos Storage;
2. a libstorage-based tutorial which covers some flavour of the the [v0.1 testnet use cases](https://docs.google.com/spreadsheets/d/1V94fhGxwTGbyLy2u8OcZmJQIzIctIGUHHfqyyY01V3E/edit?gid=2026522431#gid=2026522431).

This will be complemented by the Logos Storage UI app and its documentation.

------

Doc Packet info:

### A. Outcome + value (required)

**Outcome (end goal):** They'll have an idea of what's possible to do with Logos Storage; i.e., publish and share files. They'll also have an overview of what `libstorage` enables, and of its flows (albeit in a simplified manner).
**Why it matters:** the journey summarises the main use cases for Logos Storage in v0.1: create / download / delete / retrieve files using Logos Storage.

### B. Scope + ownership

**Journey name:** Use the Logos Storage module API (create / download / delete / retrieve) to manage files on a local storage-enabled node (though this is not the Logos Storage module API yet, it's libstorage).
**Owner (GitHub + Discord):** @gmega 
**Applies to:** https://github.com/logos-storage/easylibstorage, commit `36b0242`.
**Runtime target:** local (could be easily modified to run on the Logos testnet v0.1).

### C. Runnable happy path

The tutorial describes all of this.

### D. Configuration

Again, all described in the tutorial.

### E. Hardware requirements (required for node/client journeys)

Should work on any x86_64 or arm64 computer. Needs minimal RAM (e.g. 4GB), minimal disk space (e.g. 100 MB).

### F. Verify + troubleshoot

Included in the tutorial.

### G. Limits for v0.1

None to report (for now).

### H. References

None yet, will probably put together architecture docs over time but those are not there yet.

### Optional (only if you can)

* Takes 10-30 mins to complete (depending on expertise).
* Nothing noteworthy in terms of security/safety.

PR #459, opened 2026-08-19
docs(storage): new storage API tutorial

gmega:
This supersedes #166 with a more modern way of building a "CLI-like" app in Logos as a wrapper module that uses Storage. It addresses several of @fryorcraken's concerns, including:

1. no longer mentions libstorage or any of its derivatives: module is the sole focus;
2. no longer requires an external app skeleton: uses the (now available and consolidated) Logos toolchain;
3. deals with NAT transparently, because now we're able to;
4. (hopefully) resolves the ambiguity of what's meant by a "CLI app" in this context.

That said, the tutorial does not exercise **the whole** API as that would make it enormous. If we feel that's needed, I can do a second tutorial on streaming.


docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,513 @@
...
+  if (!started.success) {
+    return started;
+  }
+
+  // Storage has its own internal op timeout so I don't need one here.
+  const std::string result = g_result.get_future().get();
+  const json payload = parsePayload(result);
+  return StdLogosResult{.success = payload.value("success", false),
+                        .value = payload,
+                        .error = ""};
```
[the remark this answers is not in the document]
gmega: I don't think this has to be this perfect.

weboko (commented):
Dogfooded this end to end on a clean Ubuntu 24.04 container — installed Nix from scratch, ran every step, and built and loaded the module against a real `logoscore` daemon. The structure is good and the approach is a clear improvement over #166; the concurrency explanation around `syncTransferOp` is genuinely the best part. But **the tutorial cannot currently be completed as written** — two hard blockers stop the build, and Vale CI is already red.

## Blockers (build fails)

1. **`flake.nix` never gets the `storage_module` input.** Step 2 only edits `metadata.json`, so Step 5 fails with `fatal error: storage_module_api.h: No such file or directory`. The module builder resolves core-module dependencies from flake inputs — its own `ui-qml-backend` template documents this.
2. **The anonymous `namespace {` is never closed.** Assembling the Step 4 listings verbatim yields a file with unbalanced braces; GCC reports `expected '}' at end of input` plus `definition of ... is not in namespace enclosing 'StorageCliImpl'`.

With just those two fixes the build is green and produces exactly the `result/logos-storage_cli-module-lib.lgx` you document.

## CI is red

`vale` and `prose` are both failing on this PR right now. I reproduced all 7 errors locally with the pinned Vale 3.11.1: six `Logos.SpellingGB` (`namespace`, `mutex`, `serialize`, `uploadUrl`, `initialize`, `reinitialize`) and one `Google.Quotes`. Inline suggestions cover each. `build` passes, and the site builds fine for me too.

## What I verified working

Steps 1, 2, 3, 4, 6, 7, 8 and 9 all work exactly as written once the blockers are fixed:

- `nix flake init -t github:logos-co/logos-module-builder/0.2.0` scaffolds as described
- `nix build '#lgx-portable'` succeeds (the bare `#attr` form is fine on Nix 2.35) and `ls result` matches your output verbatim
- `lgpd --version 2.1.2 download storage_module -o .` and both `lgpm install` calls succeed
- `logoscore -D --config-dir ./config-dir -m ./modules` starts, and `status` reproduces your Step 9 table **character for character** — module names, versions and states all match. Worth knowing: `logoscore` emits JSON when piped and the table only on a TTY, so that block is right for an interactive reader.
- `load-module storage_cli` prints your exact two lines, and the daemon loads `storage_module` as a dependency

## What I could not verify

**Steps 10 and 11 (publish/download) I could not complete**, and I do not think that is your bug: my container only allows outbound 80/443, so the node logged `Started Storage node ... addrs=[]`, never joined `logos.test`, and `storageStart` never fired. Someone on an unrestricted network should confirm the CIDs and progress output.

That said, the failure mode is worth hardening regardless of my sandbox: `gStartedFut.get()` has no timeout, so the handler thread parks forever while the CLI returns a bare `RPC_FAILED` after ~20s, and each retry leaks another blocked thread. I have suggested a bounded wait and a prerequisite noting the connectivity requirement — that combination turns the most likely reader failure into something self-diagnosing.

## Smaller items

Frontmatter needs a pass: `sidebar_position: 1` collides with `run-logos-storage-node.md`, `title` and the H1 disagree, the slug does not match the filename, and `topics:` is missing. There is also some line-numbering drift in Step 4 (`showLineNumbers=12` should be 17; "line 9" should be 13) and two sample-output mismatches (`13 of 13` bytes should be 14; the log prefix is doubled in reality). All inline.

One optional thought on scope: I would not add the streaming tutorial you mention. This one already carries a lot, and the `syncTransferOp` pattern is the transferable idea.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+      "find_packages": [],
+      "extra_sources": [],
+      "extra_include_dirs": [],
+      "extra_link_libraries": []
+    }
+  }
+}
+```
+
+Apart from the plugin's name and description, we've added `"dependencies": ["storage_module"],` to indicate that this module depends on the Storage module, and `"concurrency": "multi"` so that the Logos request handler runs the calls on a thread that is separate from the single-threaded event loop -- this will make our lives "easier" when we try to turn the asynchronous Storage API calls into synchronous ones.
```
weboko: **Blocker — the tutorial never wires `storage_module` into `flake.nix`, so Step 5 cannot build.**

Declaring the dependency in `metadata.json` alone is not enough. The generated `logos_sdk.h` starts `#include`-ing `storage_module_api.h`, but nothing puts that header on the include path, so `nix build` dies:

```
generated_code/include/logos_sdk.h:6:10: fatal error: storage_module_api.h: No such file or directory
```

`logos-module-builder` resolves core-module dependencies from **flake inputs** — its own `ui-qml-backend` template says so explicitly: `# Add core module dependencies as inputs (must match metadata.json "dependencies")`. The `minimal-module` template you init from in Step 1 has no such input, so the reader has to add it.

Adding the input is what got me to a green build. Two further things are worth pinning while you are here: the template's `logos-module-builder` input is **unversioned**, so it silently locked to today's `main` (`378bdba`) rather than the `0.2.0` you pin the template to in Step 1; and the build resolved `logos-storage-module` to `20eda67` while Step 6 installs the pinned **2.1.2** release. Pinning both keeps the tutorial reproducible, exactly as you already do for `LOGOSCORE_TAG`/`LGPD_TAG`/`LGPM_TAG`.

````suggestion
Apart from the plugin's name and description, we've added `"dependencies": ["storage_module"],` to indicate that this module depends on the Storage module, and `"concurrency": "multi"` so that the Logos request handler runs the calls on a thread that is separate from the single-threaded event loop -- this will make our lives "easier" when we try to turn the asynchronous Storage API calls into synchronous ones.

Declaring the dependency in `metadata.json` tells the Logos runtime what to load, but the build also needs the Storage module's generated API headers. The module builder takes those from your flake inputs, so you must add a matching input to `flake.nix`. Replace the `inputs` block with:

```nix
  inputs = {
    logos-module-builder.url = "github:logos-co/logos-module-builder/0.2.0";
    storage_module.url = "github:logos-co/logos-storage-module/v2.1.2";
  };
```

The input name must match the entry in `"dependencies"`. Pinning both inputs keeps your build reproducible, and pinning `storage_module` to `v2.1.2` matches the release you install in Step 6.
````

File: docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md [hunk 3]
```diff
@@ -0,0 +1,515 @@
...
+```
+
+It is worth analysing what it does: lines 76-80 wait on the `gStarted` promise, which is set when the node starts. Line 85 acquires the operation lock and effectively blocks two `syncTransferOp` calls from running concurrently beyond that point. Lines 87-89 set up the state for the operations by clearing the completion counters and resetting the result promise.
+
+Line 92 actually sends the operation to the Storage module and, if the operation dispatches successfully (line 93), it proceeds to wait on the promise that will be set by the `onDone` callback shown in the previous listing when the operation completes. Finally, once the operation completes, Lines 99-102 extract the result and return it.
+
+The final part of our implementation contains the operations themselves (which invoke the `syncTransferOp` helper above), the missing callbacks, and the implementation for the `onContextReady` hook:
+
+```cpp showLineNumbers=104
+void StorageCliImpl::onContextReady() {
```
weboko: **Blocker — the anonymous `namespace {` opened in the constants listing is never closed, so the assembled file does not compile.**

Concatenating the Step 4 listings in order gives a file with one more `{` than `}`. Because the `StorageCliImpl::` member definitions then land inside the anonymous namespace, GCC rejects it twice over:

```
error: definition of 'auto StorageCliImpl::download(...)::<lambda>::operator()() const'
       is not in namespace enclosing 'StorageCliImpl::download(...)'
error: expected '}' at end of input
note: to match this '{'   ->  line 17: namespace {
```

Closing the namespace before the member definitions is the fix — the helpers above (`echo`, `printProgress`, `parsePayload`, `onProgress`, `onDone`, `syncTransferOp`) are exactly what belongs in it. With this change plus the `flake.nix` input, the build goes green and produces `result/logos-storage_cli-module-lib.lgx` as documented.

Note this shifts the rest of the listing down by two lines, so the `showLineNumbers` start stays at 104 but the `init`/`start` line reference below needs updating too (see my comment there).

`````suggestion
```cpp showLineNumbers=104
}  // namespace

void StorageCliImpl::onContextReady() {
`````

File: docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md [hunk 4]
```diff
@@ -0,0 +1,515 @@
...
+#include <iostream>
+#include <mutex>
+#include <system_error>
+#include <nlohmann/json.hpp>
+
+using nlohmann::json;
+
+```
+
+Note that `nlohmann/json` (included in line 9) ships with the Logos SDK, so we don't need to install it separately. Next, we open our namespace and define some constants - namely, the configuration we will be using for our node and the chunk size for transfer operations.
```
weboko: **Vale CI failure + wrong line reference.**

Two things on this line:

1. `namespace` trips `Logos.SpellingGB` (`'namespace' is not in British English`). The `vale` and `prose` checks are currently **failing** on this PR; backticking it as code silences the rule correctly rather than whitelisting a code keyword in the vocabulary.
2. `nlohmann/json` is on **line 13** of that listing, not line 9 — counting the block from `// storage_cli_impl.cpp` at 1, the blank line at 4 puts `<future>` at 9 and `<nlohmann/json.hpp>` at 13.

```suggestion
Note that `nlohmann/json` (included in line 13) ships with the Logos SDK, so we don't need to install it separately. Next, we open our anonymous `namespace` and define some constants - namely, the configuration we will be using for our node and the chunk size for transfer operations.
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+
+```cpp showLineNumbers=12
```
weboko: **Line numbering restarts too early.** The previous listing ends at line 16 (`using nlohmann::json;` is 15, followed by a blank), so this block should continue at 17, not 12. Every later block in this step chains correctly (23 -> 39 -> 67 -> 74 -> 104), so this is the only break in the sequence.

````suggestion
```cpp showLineNumbers=17
````

File: docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md [hunk 6]
```diff
@@ -0,0 +1,515 @@
...
+    "nat": "auto",
+    "data-dir": "/tmp/logos-storage",
+    "network": "logos.test"
+})";
+
+constexpr int64_t kChunkSize = 65536;
+
+```
+
+The Storage module exposes mostly asynchronous API calls, and those are not convenient for CLIs in which the user expects the operation to be complete once control returns to the terminal. Our plan, therefore, is to make both `publish` and `download` fully synchronous. We will rely on standard C++ [promises](https://en.cppreference.com/cpp/thread/promise) to do that. We will also use a [mutex](https://en.cppreference.com/cpp/thread/mutex) to serialize `publish/download` operations and keep our bookkeeping simple:
```
weboko: **Vale CI failure.** `mutex` and `serialize` both trip `Logos.SpellingGB`. Backticking `mutex` inside the link text matches how you already treat `publish`/`download`, and `serialise` is the British form the repo enforces (the rest of the docs use `anonymised`, `synchronise`).

```suggestion
The Storage module exposes mostly asynchronous API calls, and those are not convenient for CLIs in which the user expects the operation to be complete once control returns to the terminal. Our plan, therefore, is to make both `publish` and `download` fully synchronous. We will rely on standard C++ [promises](https://en.cppreference.com/cpp/thread/promise) to do that. We will also use a [`mutex`](https://en.cppreference.com/cpp/thread/mutex) to serialise `publish/download` operations and keep our bookkeeping simple:
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+  if (j.is_discarded()) {
+    echo("Failed to parse payload: " + payload);
+    return {};
+  }
+  return j;
+}
+
+```
+
+We also define `onDone` and `onProgress` callbacks, which will be invoked by the Storage module's [uploadUrl](https://logos-co.github.io/logos-storage-module/latest/api_reference.html#_CPPv4N17StorageModuleImpl9uploadUrlERKNSt6stringE7int64_t) and [downloadToUrl](https://logos-co.github.io/logos-storage-module/latest/api_reference.html#_CPPv4N17StorageModuleImpl13downloadToUrlERKNSt6stringERKNSt6stringEb7int64_t) operations as they complete or make progress, respectively:
```
weboko: **Vale CI failure.** `uploadUrl` trips `Logos.SpellingGB`. Backticking it inside the link text fixes it (and `downloadToUrl` reads better the same way, even though the dictionary happens not to flag it).

```suggestion
We also define `onDone` and `onProgress` callbacks, which will be invoked by the Storage module's [`uploadUrl`](https://logos-co.github.io/logos-storage-module/latest/api_reference.html#_CPPv4N17StorageModuleImpl9uploadUrlERKNSt6stringE7int64_t) and [`downloadToUrl`](https://logos-co.github.io/logos-storage-module/latest/api_reference.html#_CPPv4N17StorageModuleImpl13downloadToUrlERKNSt6stringERKNSt6stringEb7int64_t) operations as they complete or make progress, respectively:
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+  echo("Downloading " + cid + " to " + path.string());
+
+  return syncTransferOp("download", 0, [&] {
+    return modules().storage_module.downloadToUrl(cid, path.string(), false,
+                                                  kChunkSize);
+  });
+}
+```
+
+The implementation is mostly straightforward: we use `onContextReady` to initialize the Storage module if not already, and the download and upload operations are thin wrappers on `syncTransferOp` that invoke the appropriate Storage module API and deal with input and output files.
```
weboko: **Vale CI failure.** `initialize` trips `Logos.SpellingGB`; the repo enforces British spelling.

```suggestion
The implementation is mostly straightforward: we use `onContextReady` to initialise the Storage module if not already, and the download and upload operations are thin wrappers on `syncTransferOp` that invoke the appropriate Storage module API and deal with input and output files.
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+
+Note that we assume, in lines 121-130, that failures in dispatching `init` or `start` mean that the module has already been initialized/started. The reality is that we do not know and have no simple API to query that, so we cannot distinguish between "already initialized/started" and "failed to initialize/started". The safest approach is to not reload this module, or to reinitialize the whole node in case you do.
```
weboko: **Vale CI failures + line reference drift.** Three `Logos.SpellingGB` hits (`initialized`, `initialize`, `reinitialize`) plus a `Google.Quotes` error (the period belongs inside the closing quotation mark). The line reference also needs to become 123-132 once the anonymous namespace is closed above.

```suggestion
Note that we assume, in lines 123-132, that failures in dispatching `init` or `start` mean that the module has already been initialised/started. The reality is that we do not know and have no simple API to query that, so we cannot distinguish between "already initialised/started" and "failed to initialise/started." The safest approach is to not reload this module, or to reinitialise the whole node in case you do.
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+product: storage
+steps_layout: sectioned
+authors: giuliano
+owner: logos
+doc_version: 1
+slug: building-a-simple-logos-app-on-the-storage-module-api
+sidebar_position: 1
+---
+
+# Building a CLI app on the Storage API
```
weboko: **Frontmatter drifts from the repo conventions, and `sidebar_position` collides.**

Four things, all fixable in one edit:

- **`sidebar_position: 1` is already taken** by `run-logos-storage-node.md` in this same folder (2 and 3 are the UI and mix-network guides). Docusaurus breaks the tie by filename, so the ordering you get is accidental. `4` puts this at the end of Get started.
- **`title` and the H1 disagree** ("Building a simple Logos app on the Storage module API" vs "Building a CLI app on the Storage API"). Every other page in `docs/*/get-started/` keeps them identical.
- **Imperative mood**: siblings are "Build a Logos Core module that uses the Service Discovery API", "Run an LEZ wallet via the CLI" — not gerunds.
- **`slug` does not match the filename stem**, which is the convention everywhere else, and **`topics:` is missing** (siblings all set it).

```suggestion
title: Build a Logos module that uses the Storage module API
doc_type: procedure
product: storage
topics: storage
steps_layout: sectioned
authors: giuliano
owner: logos
doc_version: 1
slug: build-a-logos-module-that-uses-storage-module-api
sidebar_position: 4
---

# Build a Logos module that uses the Storage module API
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+- **Nix** with flakes enabled. Install from [nixos.org](https://nixos.org/download.html), then enable flakes:
+
+  ```bash
+  mkdir -p ~/.config/nix
+  echo 'experimental-features = nix-command flakes' >> ~/.config/nix/nix.conf
+  ```
+
+  Verify: `nix flake --help >/dev/null 2>&1 && echo "Flakes enabled"`
+
+- **Git**
```
weboko: **Missing prerequisite: outbound connectivity for the storage node.**

This is what stopped my run at Step 10. The storage node needs to reach the `logos.test` testnet over libp2p on non-standard TCP/UDP ports. On a host where only 80/443 egress is allowed, the node reports `Started Storage node ... addrs=[]`, the `storageStart` event never fires, and every `call` fails with an opaque `RPC_FAILED` about 20 seconds later — with nothing in the tutorial to tell you why. Worth stating up front.

```suggestion
- **Git**
- Outbound network access for the storage node. It joins the `logos.test` testnet over libp2p on non-standard TCP and UDP ports (discovery binds `0.0.0.0:8090`), so a host that only permits ports 80/443 will start the node but never finish joining, and every call will fail.
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+
+```
+
+Next, comes the implementation of our "transfer helper" - a function that handles the common logic for both publish and download operations, and is perhaps the most important part of our module:
+
+```cpp  showLineNumbers=74
+StdLogosResult syncTransferOp(const std::string &what, int64_t total,
+                              const std::function<StdLogosResult()> &op) {
+  echo("Waiting for node to start.");
+  if (!gStartedFut.get()) {
```
weboko: **`gStartedFut.get()` waits forever, and the comment below about Storage's internal timeout does not cover this wait.**

The internal timeout applies to the *operation*, not to node startup. If `storageStart` never arrives, this line blocks the handler thread indefinitely: the CLI gives up with `RPC_FAILED` after ~20s while the module thread stays parked, and because `concurrency` is `multi` every retry leaks another blocked thread. I hit this for real — three `publish` attempts each logged `Waiting for node to start.` and none ever returned.

Since this is the one place a reader is most likely to get stuck, a bounded wait turns a mystery hang into a clear message:

```suggestion
  if (gStartedFut.wait_for(std::chrono::seconds(60)) != std::future_status::ready) {
    return StdLogosResult{.success = false,
                          .value = {},
                          .error = "Timed out waiting for the storage node to "
                                   "start; check network connectivity."};
  }
  if (!gStartedFut.get()) {
```

(That needs `#include <chrono>` alongside the other headers in the first listing.)

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+    "sessionId": "0",
+    "success": true
+  }
+}
+```
+
+Because the `onProgress` callback is invoked in the module, the progress logs will show up in the daemon terminal. This is a small file, so progress is a single line:
+
+```text
+[2026-08-19 18:58:13.540] [out] [storage_cli]   100% (13 of 13 bytes)
```
weboko: **The byte count does not match the file the reader just created.** `echo "Hello, World!" > hello.txt` writes 14 bytes — 13 characters plus the trailing newline. My run logged `uploading .../hello.txt (14 bytes)`, so the progress line reads `14 of 14`.

```suggestion
[2026-08-19 18:58:13.540] [out] [storage_cli]   100% (14 of 14 bytes)
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+  }
+}
+```
+
+In the daemon logs, you should see the download progressing, e.g.:
+
+```text
+[2026-08-19 19:04:53.920] [out] [storage_cli] [storage_cli] Downloading zDvZRwzkzrrYB6sS1rRpRLt4gBhc1pWoyTSjkfszfmj1seaYYLCZ to /home/giuliano/logos-v0.2.1/./farewell-to-westphalia.pdf
+[2026-08-19 19:04:53.920] [out] [storage_cli] Waiting for node to start.
+[2026-08-19 19:04:53.920] [out] [storage_cli] Node is started.
```
weboko: **The sample daemon log is inconsistent with itself and with the code.**

Two small mismatches in this listing:

- The `[storage_cli]` prefix is doubled on the `Downloading` line above but single on these. In a real run *every* line is doubled — the daemon tags module stdout with `[storage_cli]` and `echo()` prepends its own `[storage_cli] `. My log shows `[out] [storage_cli] [storage_cli] Waiting for node to start.` for all of them.
- `Node is started.` is not a string the module can emit; `syncTransferOp` echoes `"Node is started, attempting to run " + what + " operation."`.

```suggestion
[2026-08-19 19:04:53.920] [out] [storage_cli] [storage_cli] Waiting for node to start.
[2026-08-19 19:04:53.920] [out] [storage_cli] [storage_cli] Node is started, attempting to run download operation.
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,515 @@
...
+
+```json showLineNumbers
+{
+  "name": "storage_cli",
+  "display_name": "Simple Storage CLI",
+  "version": "1.0.0",
+  "type": "core",
+  "interface": "universal",
+  "category": "example",
+  "description": "A simple CLI frontend module to Logos module",
```
weboko: Reads like a word is missing — presumably "to the Logos **Storage** module".

```suggestion
  "description": "A simple CLI frontend to the Logos Storage module",
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: Yes, this makes it smaller, and I don't think anyone will be confused by the difference.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: I get 13 bytes here - also don't think such small discrepancies are relevant.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: Not gonna add this - this is specific to your containerized setup.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: Funny thing that the spellchecker didn't complain about this one...

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: Forgot about this.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: Yep, forgot to paste the namespace braces from the actual code.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: Added to the previous snippet so line numbers don't change.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: Yeah forgot to update those after the header updates...

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: Also gonna leave this one out as I'm running out of time to work on this. Logs should show you something is not going right anyway.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,529 @@
...
+- **Nix** with flakes enabled. Install from [nixos.org](https://nixos.org/download.html), then enable flakes:
+
+  ```bash
+  mkdir -p ~/.config/nix
+  echo 'experimental-features = nix-command flakes' >> ~/.config/nix/nix.conf
+  ```
+
+  Verify: `nix flake --help >/dev/null 2>&1 && echo "Flakes enabled"`
+
+- **Git**
```
weboko: The three CLIs are shipped as AppImages, which require FUSE at runtime. On a fresh Ubuntu 24.04 (which the Prerequisites list as a supported OS), running any of `logoscore`, `lgpd`, or `lgpm` before installing `libfuse2` fails with:

```
Error: No suitable fusermount binary found on the $PATH
```

Adding it to the prereqs saves the reader a detour.

```suggestion
- **Git**
- **libfuse2** (Linux only) — the `logoscore`, `lgpd`, and `lgpm` releases below are AppImages and need `fusermount` at runtime. On Debian/Ubuntu: `sudo apt-get install libfuse2`. Without it the tools abort with `Error: No suitable fusermount binary found on the $PATH`.
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,529 @@
...
+
+Apart from the plugin's name and description, we've added `"dependencies": ["storage_module"],` to indicate that this module depends on the Storage module, and `"concurrency": "multi"` so that the Logos request handler runs the calls on a thread that is separate from the single-threaded event loop -- this will make our lives "easier" when we try to turn the asynchronous Storage API calls into synchronous ones.
+
+You should also delete the generated `src/minimal_impl.{h,cpp}` as we will be rolling out our own:
+
+```bash
+rm src/minimal_impl.{h,cpp}
+```
+
+Next, you will need to modify your generated `flakes.nix` to include the storage module as an input:
```
weboko:
```suggestion
Next, you will need to modify your generated `flake.nix` to include the storage module as an input:
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,529 @@
...
+``` bash
+$ logoscore --config-dir ./config-dir load-module storage_cli
+
+Loaded module: storage_cli (v1.0.0)
+  Dependencies loaded: storage_module
+```
+
+## Step 10: Publish a file
+
+Let's use the CLI to publish a file:
```
weboko: Following the tutorial as-is, the very first `publish` call reliably returns `Error: callModuleMethod('storage_cli','publish') RPC call failed.` — because `onStorageStart` only fires once the storage node has finished bootstrapping (on my run, ~2 min 13 s after `Started Storage node` showed up in the daemon log; the module keeps blocking on `gStartedFut.get()` inside `syncTransferOp` until then). The RPC times out client-side long before that. The doc's expected happy-path output doesn't warn about this, so the reader thinks something is broken.

```suggestion
Let's use the CLI to publish a file. The first call after starting the daemon may return `Error: callModuleMethod('storage_cli','publish') RPC call failed.` while the storage node finishes bootstrapping against the testnet — this can take a couple of minutes. Wait until the daemon log prints `[storage_cli] Node started with result: {"message":"","success":true}` and retry the same command:
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,529 @@
...
+    "sessionId": "zDvZRwzkzrrYB6sS1rRpRLt4gBhc1pWoyTSjkfszfmj1seaYYLCZ",
+    "success": true
+  }
+}
+```
+
+In the daemon logs, you should see the download progressing, e.g.:
+
+```text
+[2026-08-19 19:04:53.920] [out] [storage_cli] [storage_cli] Downloading zDvZRwzkzrrYB6sS1rRpRLt4gBhc1pWoyTSjkfszfmj1seaYYLCZ to /home/giuliano/logos-v0.2.1/./farewell-to-westphalia.pdf
```
weboko: The current runtime prefixes the line with `[storage_cli]` once (the daemon adds one prefix; `echo("Downloading ...")` in the module doesn't add another). The sample log has the tag duplicated.

```suggestion
[2026-08-19 19:04:53.920] [out] [storage_cli] Downloading zDvZRwzkzrrYB6sS1rRpRLt4gBhc1pWoyTSjkfszfmj1seaYYLCZ to /home/giuliano/logos-v0.2.1/./farewell-to-westphalia.pdf
```

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,529 @@
...
+[2026-08-19 19:04:53.920] [out] [storage_cli] Waiting for node to start.
+[2026-08-19 19:04:53.920] [out] [storage_cli] Node is started.
```
weboko: This should match what `syncTransferOp` actually prints for the download op (see line 302 of this diff — `"Node is started, attempting to run " + what + " operation."`). What the reader will see in the daemon log is:

```suggestion
[2026-08-19 19:04:53.920] [out] [storage_cli] Node is started, attempting to run download operation.
```

weboko (commented):
Dogfooded end-to-end on a fresh Ubuntu 24.04 environment (linux/x86_64) following the tutorial literally. Everything from `nix flake init` through building the module, installing both modules, starting the daemon, `load-module storage_cli`, and `publish ./hello.txt` worked and matched the expected output. Two things were smooth surprises worth mentioning:

- `nix build '#lgx-portable'` works as printed (nix accepts the bare `#attr` fragment). The `logos-storage_cli-module-lib.lgx` shows up under `result/` exactly as the doc says.
- `logoscore --config-dir ./config-dir status` and `load-module storage_cli` produce the exact output shown in Steps 9.

What did *not* match the doc, and would trip a reader following it:

1. **The first `publish` call fails with an RPC timeout while the node bootstraps.** On my run, `onStorageStart` did not fire until ~2 min 13 s after `Started Storage node` in the daemon log (it waits until bootstrap peers time out and the storage node reports `success:true`). During that window, `logoscore ... publish` returns `Error: callModuleMethod('storage_cli','publish') RPC call failed.` — a state the doc's expected output does not describe. Retrying once the daemon log prints `Node started with result:` succeeds. Suggestion inline.
2. **`flakes.nix` → `flake.nix` typo** in Step 2. Small but breaks copy/paste and search.
3. **libfuse2 prerequisite is missing.** The three CLIs are AppImages; on a minimal Ubuntu 24.04 (a "supported OS") they refuse to run with `Error: No suitable fusermount binary found on the $PATH` until `libfuse2` is installed. Suggestion adds it to the Prerequisites block.
4. **Daemon-log samples in Step 11 drift from the current runtime**: `Node is started.` should read `Node is started, attempting to run download operation.` (matches the code the tutorial itself just introduced at line 302), and the `Downloading …` line has a duplicated `[storage_cli]` tag that does not appear in current logs.

Not posting as a blocker, but I could not verify Step 11 end-to-end from this environment — all six Storage testnet bootstrap peers (`47.76.178.164:8070`, `47.76.168.186:8070`, `34.123.182.254:8070`, `34.70.60.201:8070`, `129.212.221.44:8070`, `178.128.140.206:8070`) were unreachable, so `downloadToUrl` returned `Failed to start download`. Upload worked because it does not require reachable providers. That looks like an environmental restriction on my end rather than anything in the PR.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
gmega: The node definitely does not take 2 minutes to bootstrap - it's seconds. And publish doesn't need bootstrap to be done anyway. Something else is going on here.

weboko: ok, I need to check it in another env. I suspect this run was in VPS that had weird network setup

weboko (commented):
Ran this tutorial end to end on Linux (Nix flakes, `logoscore` 0.2.2, `lgpd`/`lgpm` 0.2.1, `logos-module-builder` 0.2.0, `storage_module` v2.1.2). It works: the module builds, both modules install, the node starts and gets relay reservations on `logos.test`, `publish ./hello.txt` returns a CID, and `download` of the book CID produced a 2,276,462-byte PDF that is byte-identical to the copy on logos.co. Unload/reload of `storage_cli` also survives and `publish` still works afterwards, so the Step 4 caveat is conservative rather than wrong.

A few things to fix before merge:

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,529 @@
...
+The implementation is mostly straightforward: we use `onContextReady` to initialise the Storage module if not already, and the download and upload operations are thin wrappers on `syncTransferOp` that invoke the appropriate Storage module API and deal with input and output files.
+
+Note that we assume, in lines 128-137, that failures in dispatching `init` or `start` mean that the module has already been initialised/started. The reality is that we do not know and have no simple API to query that, so we cannot distinguish between "already initialised/started" and "failed to initialise/started." The safest approach is to not reload this module, or to reinitialise the whole node in case you do.
+
+## Step 5: Building your module
+
+Once you have implemented the module, you can build and run it using the Logos tooling suite. To build your module, run:
+
+```bash
+nix build '#lgx-portable'
```
weboko: Should be `nix build '.#lgx-portable'`. With an empty flake reference Nix searches *upwards* from the current directory for a `flake.nix`, so a reader who has a flake anywhere in a parent directory silently builds the wrong one. Every other `nix build` in the repo uses the `.#` form.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
```diff
@@ -0,0 +1,529 @@
...
+```bash
+lgpm install --file ./result/logos-storage_cli-module-lib.lgx --modules-dir ./modules
+```
+
+## Step 8: Start the Logos daemon
+
+Start a new terminal, and run:
+
+```bash
+logoscore -D --config-dir ./config-dir -m ./modules
```
weboko: In a new terminal this fails with `logoscore: command not found`. The prerequisites put the tools in `$PWD/bin` and export PATH only in that shell, and Step 1 then `cd`s into `storage_cli/` — so the new terminal has neither the PATH entry nor the right working directory (`./config-dir` and `./modules` are relative to `storage_cli/`). Worth spelling out the `cd` and the `export PATH="../bin:$PATH"` here.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
weboko: Re-ran this on a clean Linux host with the pinned versions and can confirm @gmega is right — no delay and no `RPC_FAILED`. `load-module storage_cli` at 21:23:50, `Node started with result: {"message":"","success":true}` at 21:23:52, and `publish ./hello.txt` returned a CID first try. So this is not a bootstrap-time issue and the doc needs no warning here; my earlier report was the VPS network setup. Withdrawing the suggestion.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
weboko: Correcting myself here — this suggestion is wrong on the package, and I'm withdrawing it.

`libfuse2t64` ships no binaries at all (`dpkg -L libfuse2t64 | grep bin/` → 0), so installing it does not provide `fusermount`. On Ubuntu 24.04 `/usr/bin/fusermount` is a symlink to `fusermount3` from the `fuse3` package, which is already present on a normal install.

What actually happened on my host is narrower than I reported: it has no `/dev/fuse` at all, and my shell had `APPIMAGE_EXTRACT_AND_RUN=1` set, which is why the three CLIs ran fine for me the whole way through the tutorial. Without that variable they stop at `Cannot mount AppImage`. So this is a container/VM issue, not something a reader on a normal desktop will hit — same call you already made on the connectivity note.

Leaving it here only as a pointer for anyone who does hit it: `export APPIMAGE_EXTRACT_AND_RUN=1` needs no packages and works with no FUSE at all. Your call whether that is worth a troubleshooting line; no change needed for the tutorial to be correct.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
weboko: `sidebar_position: 4` landed in c9ab20e, so the collision is gone. One item from this thread is still outstanding and it is the one a reader actually sees:

The front matter `title` ("Building a simple Logos app on the Storage module API") and the H1 ("Building a CLI app on the Storage API") name the tutorial differently, so the sidebar and search result won't match the page heading. 64 of the 68 docs in the repo keep the two identical, including every other page under `storage/get-started/`. Given the PR description sets out to pin down what "CLI app" means here, having two names for the page works against that.

The slug/`topics` points are cosmetic — happy to drop those.

docs/storage/get-started/build-a-logos-module-that-uses-storage-module-api.md
weboko: Yeah, don't listen to all what AI says

weboko (approved):
lgtm and I am pretty confident in the last run of review

PR #460, opened 2026-08-20
fix(blockchain): correct "boostrapping" typo in Cryptarchia concept

devorun:
The Cryptarchia concept page misspells **bootstrapping** as "boostrapping" when it introduces the bootstrapping fork choice rule. The same paragraph already refers to and links the **bootstrapping** rule with the correct spelling, so this is a clear typo.

### Changes
- `docs/blockchain/concepts/about-cryptarchia.md`: `boostrapping` → `bootstrapping`.
- `.github/styles/config/vocabularies/Logos/accept.txt`: remove the `boostrapping` entry. The misspelled form was whitelisted in the Vale vocabulary, which is why the linter never flagged the typo. Removing it prevents the same misspelling from slipping through again — `bootstrapping` is a valid word and is already used across the docs without a vocab entry.

Single, focused fix per the contributing guidelines.


weboko (commented):
Dogfooded this end-to-end on a clean container rather than just eyeballing the diff. It holds up — no changes requested.

**What I verified**

1. **The typo is the only one.** `grep -rn boostrap` across the whole repo (excluding `.git`) returns zero hits after the change, so the vocabulary entry is not covering any other occurrence and removing it leaves nothing orphaned.

2. **The stated rationale is correct — I confirmed the causal link.** I was initially sceptical, because `.vale.ini` sets `Vale.Spelling = NO` and `Google.Spelling = NO`, which made it look like the `accept.txt` entry was inert and the removal purely cosmetic. It isn't. The repo has a custom `Logos/SpellingGB.yml` (`extends: spelling`, `dictionaries: [en_GB]`, `level: error`) that consumes the same vocabulary. Reproduced both directions on this branch:

   - typo reintroduced, `accept.txt` entry removed (this PR's state) → `57:108 error 'boostrapping' is not in British English. Logos.SpellingGB` — **1 error**
   - typo reintroduced, `accept.txt` entry restored (pre-PR state) → **0 errors**

   So the whitelist really was the reason CI never caught this, and removing it genuinely does turn a recurrence into a hard CI error. Good catch, and the second half of the change is the part that actually matters.

3. **Nothing else regresses.** Vale 3.11.1 (the version pinned in `vale.yml`) after `vale sync`:
   - changed file: `0 errors`, 9 warnings, 41 suggestions — all pre-existing `Google.Will` / `Google.Passive` / `Google.Contractions` noise on untouched lines, and `filter_mode: added` means they don't gate the PR anyway
   - full `docs/` tree at `--minAlertLevel=error`: `0 errors`

4. **Build and rendered output.** `npm ci && npm run build` succeeds. In `build/blockchain/concepts/about-cryptarchia/index.html` the misspelling is gone and the sentence renders as "the bootstrapping fork choice rule".

5. **The adjacent link still resolves.** The paragraph's `[bootstrapping](../../get-started/glossary.md#bootstrapping)` compiles to `href=/get-started/glossary#bootstrapping`, and `build/get-started/glossary/index.html` contains `id=bootstrapping` (glossary L41, `### Bootstrapping` — and its definition, "the process a Logos Blockchain or Delivery node undergoes when first connected to the network", is the right sense for this paragraph, not the Delivery peer-exchange sense at L336).

6. **No sibling typos hiding in `accept.txt`.** Since this PR establishes that whitelist entries can mask real errors, I checked the rest of the file for the same class of problem. The suspicious-looking entries are all legitimate and genuinely in use: `spel` is the SPEL CLI (`logos-co/spel`, 36 hits), `nimbos` is a real repo in `logos-ecosystem-repositories.md`, `libcalc`/`logoscore`/`TOKENA` are real symbols in the C++ and LEZ guides. Nothing else to clean up here.

**One thing for a maintainer, unrelated to the diff**

CI has not run on this PR at all — `check-runs` for `91f360d2` returns `total_count: 0`, and the PR sits at `mergeable_state: unstable` purely because of that, not because anything failed. This is a fork PR (`devorun/logos-docs`) from a `FIRST_TIME_CONTRIBUTOR`, so GitHub is holding the workflows behind the **"Approve and run workflows"** button. Someone with write access needs to click it before Vale and Build Check will report. My local runs above are a stand-in, but the real checks should be allowed to run before merge.

Nice, tightly scoped fix — it removes the typo *and* the reason the typo survived.

PR #469, opened 2026-08-31
docs(basecamp): serve the swap walkthrough at a short URL

danisharora099:
## What

`docs.logos.co/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp` is not a URL anyone can say out loud or type. The atomic swap walkthrough now serves from:

| Route | Result |
| --- | --- |
| `/basecamp/atomic-swaps-poc` | the page (canonical) |
| `/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp` | redirects to it |

## How

- **Frontmatter `slug`** on `docs/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp.md` moves the page to the short route. The file name is unchanged, so the doc ID, the sidebar position and the "Edit this page" link all still resolve.
- **`@docusaurus/plugin-client-redirects`** (already configured in this repo for the legacy GitBook paths) keeps the long path alive. The old URL is spoken in a published demo video and embedded in in-app feedback links, so it can never 404. The plugin's redirect page appends `location.search` and `location.hash`, so a deep link like `…#step-2-set-up-your-accounts` still lands on that step.
- **The sidebar link** in `sidebars.ts` was the only internal reference to the page; it now points at the short route. It has to — sidebar `href`s are checked against real routes under `onBrokenLinks: throw`, and a redirect stub is a `postBuild` file rather than a route. No markdown page cross-links to this doc.

### Why `atomic-swaps-poc` and not `swaps`

"Swap" already means two things in these docs: the cross-chain atomic swap this page covers, and the AMM swaps documented under `lez/` and `build-an-app/`. The `-poc` suffix also marks the app's current status, so the route won't need renaming away from a bare `swaps` once the app is more than a proof of concept.

## Verification

Against the production build (`npm ci && npm run build`, then `docusaurus serve`):

- `/basecamp/atomic-swaps-poc` → `200`, renders the walkthrough.
- `/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp` → `200`, serves the redirect stub pointing at `/basecamp/atomic-swaps-poc` with search and hash carried across.
- `/basecamp` index → `200`.
- Build passes with `onBrokenLinks: 'throw'`.
- `vale docs/basecamp/swap-eth-and-lez-tokens-in-logos-basecamp.md` → 0 errors (CI lints `docs` only, `filter_mode: added`).

`AGENTS.md` gains a short **Public URLs and renames** note so a future change doesn't quietly drop the redirect or repoint the sidebar at a stub.

https://claude.ai/code/session_01HUebWmDDiZbZDtASYbq9bh

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,6 +17,10 @@ Several pages under `docs/` (notably `docs/basecamp/swap-eth-and-lez-tokens-in-l
 
 For the swap page specifically, ground truth lives in `logos-co/eth-lez-atomic-swaps` at the `swap_ui-vX.Y.Z` tags: `swap-ui/src/qml/SetupSteps.js` derives the Setup step order, titles, numbering and page subtitle for both onboarding flows, `swap-ui/src/qml/SetupView.qml` holds the card copy, and `docs/DEVELOPMENT.md` documents which flow ships by default. Read those at the tag rather than the repo's working tree, since `master` can be ahead of the released catalogue build.
 
+## Public URLs and renames
+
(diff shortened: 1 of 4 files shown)


weboko (approved):
I'll try to run my agent but lgtm

PR #477, opened 2026-09-11
Blockchain FAQ first draft

kashepavadan:


docs/blockchain/faq-logos-blockchain.md
```diff
@@ -0,0 +1,45 @@
...
+
+Transaction privacy needs different vastly for different use cases, and is therefore not provided directly by Bedrock. [Notes](../get-started/glossary.md#note) and their transfers are recorded openly on the ledger, similar to Bitcoin, and the Bedrock layer makes no attempt to hide that data. Developers that need to hide their app's data, such as balances or message contents, must implement their desired level of transaction privacy at the [Zone](concepts/about-zones.md) layer instead of relying on Bedrock for it. **Transaction privacy is supported natively on the [LEZ](../get-started/glossary.md#lez), in addition to public transactions**.
```
davidrusu, line 23:
```suggestion
Transaction privacy needs differ vastly for different use cases, and is therefore not provided directly by Bedrock. [Notes](../get-started/glossary.md#note) and their transfers are recorded openly on the ledger, similar to Bitcoin, and the Bedrock layer makes no attempt to hide that data. Developers that need to hide their app's data, such as balances or message contents, must implement their desired level of transaction privacy at the [Zone](concepts/about-zones.md) layer instead of relying on Bedrock for it. **Transaction privacy is supported natively on the [LEZ](../get-started/glossary.md#lez), in addition to public transactions**.
```

docs/blockchain/faq-logos-blockchain.md
```diff
@@ -0,0 +1,45 @@
...
+
+:::note
+For more information on proposer privacy on Logos, see [Anonymous Block Proposers](https://blog.logos.co/article/anonymous-block-proposers).
+:::
+
+## Does the Blend Network always protect proposer privacy for everyone?
+
+No, that protection depends on whether the proposer is a [core node](../get-started/glossary.md#core-node), and how many other core nodes are active in a given [epoch](../get-started/glossary.md#epoch). Only declared core nodes relay and blend messages, and their participation in Blend activity makes it hard to distinguish when they publish their own proposals. A block proposer that hasn't declared as a core node participates as an edge node instead, sending its proposal through core nodes for blending - an unusual activity that can be singled out and connected to published blocks by some attackers. If too few core nodes are active, all proposers fall back to broadcasting directly, with no Blend privacy for that epoch.
```
davidrusu, line 31: This changes with EmPoWering.

With EmPoWering, blend is used not just for block proposals, but for transactions as well.

This makes submitting a message to blend not an exceptional event anymore. Block proposals _blend in_ (pardon the pun) with transactions now.

So, best to give a two tier privacy here:
1. Anonymity set for proposers that are core nodes are all core nodes + all transactions concurrent transactions
2. Anonymity set for proposers that are non-core nodes is all concurrent transactions

Nodes who want to maximize proposer privacy should join the Blend Network as a core node.

docs/blockchain/faq-logos-blockchain.md
```diff
@@ -0,0 +1,45 @@
...
+
+On the LEZ, transactions are verified differently depending on if they're public or private. Public transactions are re-executed by LEZ validators, while private transactions involve the generation of zero knowledge proofs via Risc0, which are then verified by LEZ validators. 
+
+## Does decentralised sequencing eliminate MEV on a Zone?
+
+No, it distributes the opportunity for MEV extraction across a set of sequencers rather than removing it. When a Zone uses several sequencers, taking turns on a round-robin schedule or competing to publish first, no single sequencer can reliably capture all of the MEV available the way a single-sequencer rollup's operator could. A threshold of sequencers can also jointly sign a change to the sequencer list, letting an honest majority remove a malicious one.
```
davidrusu, line 41: Add a note that by isolating MEV to Zones, we keep the Blockchain neutral and decentralized, Centralization due to MEV in one Zone does not impact the centralization in other Zones.

docs/blockchain/faq-logos-blockchain.md
kashepavadan: Thanks! Some questions:

1. What is the benefit of passing transactions through Blend for the user submitting the transaction? The transaction content is still public, right? I understand that more participants in Blend increases the anonymity set etc.

2. How can we be confident that all/most transactions will go through Blend, given that the PoW imposes an additional cost?

davidrusu: 1. transaction is not necessarily transparent, e.g. suppose you mined the the tokens required to pay for the tx and the contents of the tx is an encrypted inscription.
2. we cannot, but we can provide clients that default to using the public path. Its the same story for block proposers, they are not forced to use blend, but they benefit from it

PR #355, opened 2026-06-24
docs(lez): Extensions section with admin-authority and extension-library guides

mmlado:
## Summary

Adds two pages under a new `docs/lez/extensions/` folder, plus the section's `_category_.json`:

- [`admin-authority.md`](docs/lez/extensions/admin-authority.md), integration guide for app developers. Covers annotation, gating, AdminCandidate (Signer / PDA), initialization window risk, transfer, renounce, IDL verification, security notes, and embedded mode (the admin slot inside a consumer account via `#[admin_slot]` and `#[admin_initialize]`).
- [`build-a-spel-extension-library.md`](docs/lez/extensions/build-a-spel-extension-library.md), author guide for the generic `[package.metadata.spel]` discovery mechanism. Covers two-crate layout, metadata fields, marker / gate macros, multi-extension stacking, auto-wrap, embedded mode with bound args and slot markers, common troubleshooting.

## Context

These pages document the user-facing surface of RFP-001 (admin-authority library) and the framework mechanism it relies on. The framework mechanism, a generic `[package.metadata.spel]` extension scanner in `spel-framework-core::idl_gen`, is delivered in logos-co/spel#233. These docs are the consumer-facing complement.

Future extension libraries (e.g. RFP-002 freeze-authority) can ship without any framework change, so the "Build a SPEL extension library" page is written generically to serve as the canonical author guide for any future extension.

## Why a new `extensions/` folder

`lez/` already groups capabilities by folder (`get-started/`, `transfer-tokens/`). Extension libraries are a new capability class. Placing both pages under `lez/extensions/` mirrors the existing structure and keeps consumer and author guides discoverable side-by-side.

## Companion repositories

- [`mmlado/spel-admin-authority`](https://github.com/mmlado/spel-admin-authority): the admin-authority library implementation referenced by `admin-authority.md`.
- logos-co/spel#233: the framework PR carrying the generic extension scanner referenced by both guides.

## Open questions for maintainers

1. **Folder name.** Used `extensions/` for brevity. Alternatives: `extension-libraries/`, `program-extensions/`, `access-control/` (too narrow for the generic guide).
2. **Page placement.** Both under `lez/`. The author guide could arguably sit under `build-an-app/` instead. Open to splitting if you prefer.
3. **Cross-linking.** The pages reference each other directly. No links from the broader `lez/` pages yet, happy to add.

## Style notes

[...]

Issue #257, opened 2026-05-12
[Follow-up] Preconfigured accounts


cheny0:
Add this information to the relevant documents. Reference: https://github.com/logos-co/logos-docs/pull/249

> [!NOTE]
>
> Once initialized, `wallet account ls` (used later in Step 5) will list four preconfigured accounts (`Preconfigured Public/...` and `Preconfigured Private/...`) alongside any accounts you create. They ship with the wallet binary for built-in tests; you can ignore them in normal use, or use them as send-to targets when experimenting.

Issue #246, opened 2026-05-08
Empty docs/**/README.md files — keep, populate, or delete?


wozos:
## Question
The `docs/**/README.md` files at the eight locations below are empty (0 bytes beyond EOL). What is the intended state?

| File |
|------|
| `docs/README.md` |
| `docs/_shared/README.md` |
| `docs/apps/README.md` |
| `docs/blockchain/README.md` |
| `docs/connect/README.md` |
| `docs/core/README.md` |
| `docs/messaging/README.md` |
| `docs/storage/README.md` |

## Options
1. **Populate** — short orientation block per directory pointing at the journeys it contains.
2. **Delete** — remove the empty files; let the top-level README serve as the only landing page.
3. **Keep** — empty placeholders are intentional.

## Why this matters
A first-time visitor landing at, e.g., `https://github.com/logos-co/logos-docs/blob/main/docs/blockchain/README.md` sees a blank page, which reads as "this section is unfinished" even when the journey docs underneath are reasonably complete. The visibility cost is highest for the section landing pages (`docs/blockchain/README.md`, `docs/messaging/README.md`, etc.).

Happy to follow up with a populate-or-delete PR once the team picks a direction.

---
Source: dogfooding run on `logos-co/logos-docs@c72fda5` (Raspberry Pi 5, linux/aarch64, 2026-05-08). Finding F-15 in the local report.

weboko:
essentially the question is whether we want to populate empty `README.md` files? 

cc @fryorcraken @cheny0

cheny0:
These files were planned for the index and intro for each section (needed when there is a documentation site). We haven't got the time to populate them.

weboko:
@cheny0 should I create some AI written stub or let's keep it empty for now? cc @fryorcraken

cheny0:
Let's keep them empty for now. I'll think about what to do with them. Thanks!

cheny0:
Addressed by https://github.com/logos-co/logos-docs/pull/301

PR #478, opened 2026-09-12
docs: extension pages follow the upstream framework pin

mmlado:
Follow-up to #355, as promised there: logos-co/spel#257 merged on 2026-09-11 and the three libraries released on that commit (spel-authority v0.1.1, admin-authority v0.1.3, freeze-authority v0.1.4). The pages move with them: framework at logos-co/spel 8183b01, the two library tags, lee_core v0.2.4. No fork reference remains. The framework pin stays a rev until upstream cuts a release that contains #257, then it becomes a tag.

Every sentence tied to the old revision was re-checked at the new one:

- The CLI now links pcsc through LEZ's wallet crate, so libpcsclite-dev is a prerequisite. The documented cargo install command was run at this revision on 1.94.0.
- The circuits artifact is still v0.5.3, both LEZ tags pin the same lbc-*-sys release.
- spel init defaults to a floating pull request head, not main. The scaffold path was walked with the page's source flags: pins match, fetch, marker, host check and IDL trio all pass.
- The witness exchange (#246) is in this revision. Both transfer sections describe the co-signer flow from the CLI README and say the page has not walked it live.
- The verification sentences now name the revision each claim was checked on.

Docusaurus build green. Vale runs in CI.

### DIFF
--- a/docs/lez/extensions/admin-authority.md
+++ b/docs/lez/extensions/admin-authority.md
@@ -16,7 +16,7 @@ sidebar_position: 1
 :::warning
 This page is an early draft and may be incomplete or incorrect. Expect changes, missing prerequisites, and commands that might not work in your setup. This content is still being completed and verified.
 
-This page tracks unreleased code. The dependency snippets pin a personal fork of the framework and pre-release library tags. The pins move to logos-co sources once the extension mechanism lands upstream ([logos-co/spel#257](https://github.com/logos-co/spel/pull/257)).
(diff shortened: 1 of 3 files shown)


kashepavadan:
@mmlado Style checks are failing, mind fixing them? I can't since the PR is from a branch on your own fork of the repo.

mmlado:
Fixed. The en_GB dictionary rejects the possessive README's, and the line only started blocking because this PR rewrote it, so filter_mode: added picked up a sentence that has been there since #355. Reworded to avoid the possessive rather than growing the vocabulary. Vale is clean at error level on all three pages locally, and the build passes.

kashepavadan (approved):

kashepavadan:
Thanks. Please submit a PR when the relevant release comes out updating this!

PR #481, opened 2026-09-18
Blockchain concept docs PoW update

kashepavadan:


weboko (commented):
Reviewed and dogfooded on this branch (`361e0d4`). Everything below is verified locally, not read off the diff.

**What I ran**

- `npm ci && npm run build` - **passes**.
- `vale 3.11.1` (the CI version) at `--minAlertLevel=error` over all seven changed docs - the only error is `decentalisation` at `faq-logos-blockchain.md:45`, which is **pre-existing** and on a line this PR doesn't touch, so the `filter_mode: added` job will stay green. Worth fixing in passing since you're in that file.
- **Anchor check.** `onBrokenAnchors` isn't set in `docusaurus.config.ts`, so Docusaurus defaults to `warn` and a bad anchor would ship silently. I extracted every heading `id` from the built HTML and checked all 12 anchors this PR introduces or links to (`#mining`, `#empowering`, `#mining-notes`, `#sending-mantle-transactions-via-proof-of-work`, `#epoch`, `#channel`, `#inscription`, `#blend-network`) - **all 12 resolve**.
- Rendered the three new sections in Chromium and read them as a reader; glossary ordering is correct in both the E and M sections.

**Findings**

The structure is good and the new material reads well. The issues are two grammar slips that show up in the rendered page, a rename that makes one sentence circular, a terminology change that contradicts the rest of the docs, and one technical point I can't resolve from the docs alone (the two threshold derivations on `about-mantle.md:43` and `about-the-blend-network.md:81`).

I applied all the suggestions below locally and re-verified: `npm run build` passes, Vale is clean apart from the pre-existing typo, and all anchors still resolve - including the two new links the suggestions add.

Only the inline question on `about-the-blend-network.md:81` and the request for a source on the `10 claims per block` figure need input from you; the rest are mechanical.

docs/blockchain/concepts/about-the-blend-network.md
```diff
@@ -68,11 +69,19 @@ Participation in the Blend Network as a Blend node is more complex than running
...
 - **Broadcast**—a fallback method used when fewer than the minimum number of core nodes are active in an epoch. With too few core nodes to blend through, a proposer will broadcast its proposal directly, without Blend privacy for that epoch.
 
 Only the **core** role requires operator action—an SDP declaration, a locked stake, and a publicly reachable Blend port (see [Join the Blend Network as a core node](../blend/join-the-blend-network-as-a-core-node.md)). By contrast, a Logos Blockchain node participates as an edge node if the chain is online and enough core nodes exist, and falls back to broadcast mode otherwise. The minimum number of core nodes needed for edge participation is set by the network.
 
+### Sending Mantle transactions via Proof of Work
+
+In addition to block proposals, nodes can also send Mantle transactions directly through the Blend Network. Using a Proof of Work mechanism known as *[EmPoWering](../../get-started/glossary.md#empowering)*, even a user without any notes benefit from some of the privacy provided by the Blend Network. However, as an edge node, their transaction will not have the full privacy that core nodes enjoy.
```
weboko: Subject-verb agreement: *a user ... benefit* should be *benefits*.

This one is visible in the rendered page, so it's worth fixing before merge.
```suggestion
In addition to block proposals, nodes can also send Mantle transactions directly through the Blend Network. Using a Proof of Work mechanism known as *[EmPoWering](../../get-started/glossary.md#empowering)*, even a user without any notes benefits from some of the privacy provided by the Blend Network. However, as an edge node, their transaction will not have the full privacy that core nodes enjoy.
```

docs/blockchain/concepts/about-the-blend-network.md
```diff
@@ -68,11 +69,19 @@ Participation in the Blend Network as a Blend node is more complex than running
...
+
+While Logos note transfers and most channel [inscriptions](../../get-started/glossary.md#inscription) are not private, Proof of Work Blending can be used to inscribe arbitrary data on-chain in a private way. To do this, the user should first mine some notes and submit the claim transaction via PoW through the Blend Network. Then, they can submit an inscription transaction to a new Mantle [channel](../../get-started/glossary.md#channel) through Blend, which will add the desired data to the chain. To avoiding linking this inscription with future ones, the user should then throw away the channel's signing key.
```
weboko: Two things on this line:

1. **Grammar** - *To avoiding linking* should be *To avoid linking*.
2. **Terminology** - *Proof of Work Blending* is a third name for a mechanism this PR otherwise names *EmPoWering* (glossary) and "a Proof of Work mechanism" (line 79, and `about-mantle.md`). Using the glossary term here keeps the three pages consistent and gives the reader a link back to the definition.
```suggestion
While Logos note transfers and most channel [inscriptions](../../get-started/glossary.md#inscription) are not private, *[EmPoWering](../../get-started/glossary.md#empowering)* can be used to inscribe arbitrary data on-chain in a private way. To do this, the user should first mine some notes and submit the claim transaction via PoW through the Blend Network. Then, they can submit an inscription transaction to a new Mantle [channel](../../get-started/glossary.md#channel) through Blend, which will add the desired data to the chain. To avoid linking this inscription with future ones, the user should then throw away the channel's signing key.
```

docs/blockchain/concepts/about-the-blend-network.md
```diff
@@ -49,7 +50,7 @@ Random delays are the other component of message blending. If messages are rare
 
 ## Cover traffic
 
-An important way that the Blend Network obscures network patterns is by producing indistinguishable messages within a "crowded" network. To increase this effect in an environment where proposals are relatively rare, the Blend Network also produces artificial cover messages. Cover messages do not contain any meaningful payload and are generated by Blend nodes to increase network noise and to blend in with data messages that contain real proposals.
+An important way that the Blend Network obscures network patterns is by producing indistinguishable messages within a "crowded" network. To increase this effect in an environment where proposals are relatively rare, the Blend Network also produces artificial cover messages. Cover messages do not contain any meaningful payload and are generated by Blend nodes to increase network noise and to blend in with data messages that contain real messages.
```
weboko: This rename makes the sentence circular: *data messages that contain real messages*. The pre-change text ("that contain real proposals") was meaningful - the contrast is between cover messages, which carry no payload, and data messages, which carry a real one. Now that the page covers Mantle transactions as well as proposals, "payloads" keeps the generalisation without the tautology.
```suggestion
An important way that the Blend Network obscures network patterns is by producing indistinguishable messages within a "crowded" network. To increase this effect in an environment where proposals are relatively rare, the Blend Network also produces artificial cover messages. Cover messages do not contain any meaningful payload and are generated by Blend nodes to increase network noise and to blend in with data messages that carry real payloads.
```

docs/blockchain/concepts/about-the-blend-network.md
```diff
@@ -68,11 +69,19 @@ Participation in the Blend Network as a Blend node is more complex than running
...
 
 - **Core node**—a node that has opted in through the [Service Declaration Protocol](#service-declaration-protocol) and locked the required stake. Core nodes carry out the message blending: they relay and blend messages for others, resulting in maximum privacy for their own proposals. Core nodes receive rewards for their efforts via the [SRDP](#service-reward-distribution-protocol).
-- **Edge node**—a block proposer node that has *not* declared as a core node but participates whenever there are enough core nodes in that epoch. An edge node sends its own proposals through the core nodes for blending, gaining limited proposer privacy without relaying for others. This is automatic and needs no configuration or stake.
+- **Edge node**—a node sending a message through Blend (such as a block proposal or Mantle transaction) that has *not* declared as a core node but participates whenever there are enough core nodes in that epoch. An edge node sends its own messages through the core nodes for blending, gaining limited privacy without relaying for others.
```
weboko, line 72: Two issues introduced by the rewrite:

1. The relative clause *that has \*not\* declared as a core node* now sits directly after "Mantle transaction", so it reads as if the transaction is what declares. Moving it back next to "node" fixes the attachment.
2. The rewrite drops *"This is automatic and needs no configuration or stake."* That is still true and is the one thing that distinguishes an edge node from a core node operationally - the very next paragraph leans on it ("Only the **core** role requires operator action"). Worth keeping, minus the now-inaccurate "automatic" (a PoW transaction is deliberate, not automatic).
```suggestion
- **Edge node**—a node that has *not* declared as a core node but that sends its own messages (such as a block proposal or a Mantle transaction) through Blend, participating whenever there are enough core nodes in that epoch. An edge node sends its own messages through the core nodes for blending, gaining limited privacy without relaying for others. Acting as an edge node needs no declaration and no stake.
```

docs/blockchain/concepts/about-the-blend-network.md
```diff
@@ -68,11 +69,19 @@ Participation in the Blend Network as a Blend node is more complex than running
...
 - **Broadcast**—a fallback method used when fewer than the minimum number of core nodes are active in an epoch. With too few core nodes to blend through, a proposer will broadcast its proposal directly, without Blend privacy for that epoch.
```
weboko: The **Edge node** bullet above was generalised from proposals to messages, but the **Broadcast** bullet still talks only about a *proposer* broadcasting a *proposal*. If a Mantle transaction can be sent through Blend as an edge node, the same fallback question applies to it, so the two bullets should be generalised together.
```suggestion
- **Broadcast**—a fallback method used when fewer than the minimum number of core nodes are active in an epoch. With too few core nodes to blend through, a sender will broadcast its message directly, without Blend privacy for that epoch.
```

docs/blockchain/concepts/about-the-blend-network.md
```diff
@@ -68,11 +69,19 @@ Participation in the Blend Network as a Blend node is more complex than running
...
 
+### Sending Mantle transactions via Proof of Work
+
+In addition to block proposals, nodes can also send Mantle transactions directly through the Blend Network. Using a Proof of Work mechanism known as *[EmPoWering](../../get-started/glossary.md#empowering)*, even a user without any notes benefit from some of the privacy provided by the Blend Network. However, as an edge node, their transaction will not have the full privacy that core nodes enjoy.
+
+To send a Mantle transaction through the Blend Network, a user chooses a value and hashes it together with the consensus randomness value for the current or previous epoch. This hash is compared to a threshold value derived from the transactions in the epoch before last - if the hash is below the threshold, the user is eligible to send a Mantle transaction via the Blend Network. This transaction gets added to the mempool of the last node on its Blend path, eventually being included in a block in a way that makes it infeasible to link the transaction with the user that sent it.
```
weboko, line 81: Worth double-checking against the spec: this says the threshold is *derived from the transactions in the epoch before last*, while `about-mantle.md:43` says the threshold for the same *EmPoWering* PoW is *derived from a recent canonical block selected by the user*. The two sections otherwise describe an identical procedure (choose a value, hash it with the consensus randomness of the current or previous epoch), so as written a reader cannot tell whether these are one mechanism with two thresholds or two distinct PoWs that happen to share a name.

No suggested change, because I cannot tell from the docs alone which is correct - could you confirm and then either reconcile the wording or say explicitly that the two use different difficulty targets?

docs/blockchain/concepts/about-mantle.md
```diff
@@ -36,6 +36,12 @@ Mantle maintains a restricted ledger that keeps track of fungible assets known a
...
+
+Owning notes is a prerequisite to meaningful involvement in the Logos Blockchain, allowing users to participate in consensus, join the Blend Network, and post transactions. To allow users to independently acquire notes in a private, decentralised way, Mantle supports mining notes via a Proof of Work mechanism known as *[EmPoWering](../../get-started/glossary.md#empowering)*. Notes are mined by solving a cryptographic "puzzle" - similar to how Bitcoin is mined.
+
+To mine a new note, a user chooses a value and hashes it together with the consensus randomness value for the current or previous [epoch](../../get-started/glossary.md#epoch). This hash is compared to a threshold value derived from a recent canonical block selected by the user - if the hash is below the threshold, the user is eligible to claim some notes from Mantle's reward pool. The difficulty of solving the puzzle (that is, how many tries it would take on average to find a winning value) is adjusted periodically to ensure that only 10 claims are made per block. A user with a winning hash can submit their claim transaction via the Blend Network to make it infeasible to link the transaction with their identity.
```
weboko: *adjusted periodically to ensure that only 10 claims are made per block* overstates what a difficulty retarget can do. Difficulty adjustment tracks a moving average, so per-block claim counts still vary; it targets a rate, it cannot guarantee a per-block cap. "Target an average of" says the same thing without promising a hard limit.

Also, could you add a spec or research-post link for the value 10? Per `AGENTS.md`, exact protocol constants need a ground-truth source, and there is nothing in this repo a future editor could check it against.
```suggestion
To mine a new note, a user chooses a value and hashes it together with the consensus randomness value for the current or previous [epoch](../../get-started/glossary.md#epoch). This hash is compared to a threshold value derived from a recent canonical block selected by the user - if the hash is below the threshold, the user is eligible to claim some notes from Mantle's reward pool. The difficulty of solving the puzzle (that is, how many tries it would take on average to find a winning value) is adjusted periodically to target an average of 10 claims per block. A user with a winning hash can submit their claim transaction via the Blend Network to make it infeasible to link the transaction with their identity.
```

docs/blockchain/introduction-to-the-logos-blockchain.md
```diff
@@ -38,13 +38,13 @@ The Logos Blockchain is implemented as two blockchain layers. Application execut
 
 ### Bedrock
 
-Bedrock is a large-scale validator network that serves as the foundational layer of the Logos Blockchain. It provides consensus, data availability, and lightweight verification to Zones. Its [Private Proof of Stake](../get-started/glossary.md#private-proof-of-stake) ([PPoS](../get-started/glossary.md#ppos)) consensus protocol, [Cryptarchia](concepts/about-cryptarchia.md), keeps block proposers private while staying scalable, resilient, and accessible. Bedrock also enables decentralised sequencing for Zones, token bridging, and inter-Zone messaging.
+Bedrock is a large-scale validator network that serves as the foundational layer of the Logos Blockchain. It provides consensus, data availability, and lightweight verification to Zones. Its [Private Proof of Stake](../get-started/glossary.md#private-proof-of-stake) ([PPoS](../get-started/glossary.md#ppos)) consensus protocol, [Cryptarchia](concepts/about-cryptarchia.md), keeps block proposers private while staying scalable, resilient, and accessible. Bedrock also enables decentralised sequencing for Zones, note [mining](../get-started/glossary.md#mining) & bridging, and inter-Zone messaging.
```
weboko: This replaces *token bridging* with *note ... bridging*, which conflicts with the rest of the docs. "Token bridging" is the established term: it is the heading of `concepts/about-mantle.md#token-bridging` and `concepts/about-zones.md#token-bridging`, and it is used in `about-bedrock.md` and the FAQ. Line 53 of this same file still reads "Bedrock also enables token bridging and decentralised sequencing", so the page now contradicts itself.

Also, `&` is used as a conjunction here - elsewhere in `docs/` the only ampersands in prose are inside code comments or a literal UI label ("Confirm & deposit").
```suggestion
Bedrock is a large-scale validator network that serves as the foundational layer of the Logos Blockchain. It provides consensus, data availability, and lightweight verification to Zones. Its [Private Proof of Stake](../get-started/glossary.md#private-proof-of-stake) ([PPoS](../get-started/glossary.md#ppos)) consensus protocol, [Cryptarchia](concepts/about-cryptarchia.md), keeps block proposers private while staying scalable, resilient, and accessible. Bedrock also enables decentralised sequencing for Zones, note [mining](../get-started/glossary.md#mining), token bridging, and inter-Zone messaging.
```

docs/blockchain/faq-logos-blockchain.md
```diff
@@ -20,7 +20,9 @@ It is important to distinguish between two distinct kinds of privacy that are po
...
+Transaction privacy needs differ vastly for different use cases, so this type of privacy is not provided directly by Bedrock for most cases. [Notes](../get-started/glossary.md#note) and their transfers are recorded openly on the ledger, similar to Bitcoin, and the Bedrock layer makes no attempt to hide that data. Despite this, it is possible to mine new notes and write arbitrary data on-chain in a private way by submitting these types of transactions via the [Blend Network](../get-started/glossary.md#blend-network). Bedrock therefore supports truly private consensus participation and [inscriptions](../get-started/glossary.md#inscription), but not note transfers or other transaction types.
```
weboko, line 23: The *therefore* doesn't follow from the sentence before it. That sentence establishes two private operations - **mining new notes** and **writing arbitrary data on-chain** - but this one concludes with "private consensus participation and inscriptions". Consensus participation is what Cryptarchia and Blend already gave you; it isn't what the new PoW path adds.

The trailing clause is also self-contradicting: a mining claim *is* a transaction type, so "but not note transfers or other transaction types" excludes the thing the previous sentence just said was private.
```suggestion
Transaction privacy needs differ vastly for different use cases, so this type of privacy is not provided directly by Bedrock for most cases. [Notes](../get-started/glossary.md#note) and their transfers are recorded openly on the ledger, similar to Bitcoin, and the Bedrock layer makes no attempt to hide that data. Despite this, it is possible to mine new notes and write arbitrary data on-chain in a private way by submitting these types of transactions via the [Blend Network](../get-started/glossary.md#blend-network). Bedrock therefore supports truly private note [mining](../get-started/glossary.md#mining) and [inscriptions](../get-started/glossary.md#inscription), but not note transfers or any other transaction type.
```

docs/get-started/glossary.md
```diff
@@ -98,6 +98,10 @@ A peer discovery mechanism used in Logos Delivery that allows the retrieval of a
 
 ## E
 
+### EmPoWering
+
+A term used to describe how Proof of Work (PoW) is used on the Blend Network. See [About Mantle](../blockchain/concepts/about-mantle.md#mining-notes) and [About the Blend Network](../blockchain/concepts/about-the-blend-network.md#sending-mantle-transactions-via-proof-of-work).
```
weboko: The definition is narrower than what this PR documents. `about-mantle.md#mining-notes` presents EmPoWering as the PoW used to mine notes from Mantle's reward pool - a Mantle mechanism - and the Blend Network is only where the resulting claim transaction is submitted. Defining the term as "how PoW is used on the Blend Network" doesn't cover the first of the two pages it then links to.
```suggestion
A term used to describe how Proof of Work (PoW) is used on the Logos Blockchain, both to mine notes and to send Mantle transactions through the Blend Network. See [About Mantle](../blockchain/concepts/about-mantle.md#mining-notes) and [About the Blend Network](../blockchain/concepts/about-the-blend-network.md#sending-mantle-transactions-via-proof-of-work).
```

.github/styles/config/vocabularies/Logos/accept.txt
```diff
@@ -6,6 +6,7 @@
 [Cc]ounterparty
 [Dd]elivery
 [Dd]ev
+EmPoWering
```
weboko, line 9: Optional: this entry looks unnecessary. I removed it locally and re-ran `vale --minAlertLevel=error` over `glossary.md`, `about-mantle.md` and `about-the-blend-network.md` - still clean, because `Logos.SpellingGB` doesn't flag mixed-case words like `EmPoWering`. I confirmed the check was live in the same run by adding a nonsense word, which it did flag.

Harmless to keep, so entirely up to you.

davidrusu (approved):

PR #482, opened 2026-09-18
Alisher + legal audit changes

kashepavadan:


docs/build-an-app/tutorials/build-app-with-logos-zone-sdk.md
```diff
@@ -1,16 +1,16 @@
...
+title: Build an App with the Logos Zone SDK
 doc_type: procedure
 product: blockchain
 topics: [channels, zones]
 steps_layout: sectioned
 authors: kashepavadan
 owner: logos
 doc_version: 1
-slug: build-dapp-with-logos-zone-sdk
+slug: build-app-with-logos-zone-sdk
```
weboko: **This breaks a published URL.** The rename changes the page's `slug`, and `slug` is the public URL, so `/build-an-app/tutorials/build-dapp-with-logos-zone-sdk` stops resolving the moment this merges. Nothing in this PR adds a redirect for it.

Verified against production and against a local build of this branch:

```
$ curl -s https://docs.logos.co/build-an-app/tutorials/build-dapp-with-logos-zone-sdk | grep -o '<title>[^<]*'
<title>Build a dApp with the Logos Zone SDK | Logos documentation
```

```
$ npm run build && ls build/build-an-app/tutorials/
build-app-with-logos-zone-sdk          # new
# build-dapp-with-logos-zone-sdk is gone -> 404
```

The old path is live today, and `AGENTS.md` is explicit that some published URLs live outside this repo (demo videos, in-app feedback links) and cannot be fixed after the fact, so a `slug` must never change without leaving the old path redirecting.

The build won't catch this — `onBrokenLinks: 'throw'` only checks links *inside* the site, and this PR already updated the one internal reference in `sidebars.ts`.

Please add a redirect alongside the rename, in the `redirects` array in `docusaurus.config.ts`:

```ts
{
  from: '/build-an-app/tutorials/build-dapp-with-logos-zone-sdk',
  to: '/build-an-app/tutorials/build-app-with-logos-zone-sdk',
},
```

(The redirect plugin carries the query string and hash across, so deep links to a step survive.)

docs/basecamp/install-and-load-a-module-in-logos-basecamp.md
```diff
@@ -51,35 +51,37 @@ When installing a module, Logos Basecamp extracts the variant for your platform
...
+Confirm the module actually installed: check its status in **Package Manager**.
+
+- If installing from the online catalogue: if a package's row shows **Not available** instead of **Install**, it has no build for your platform, build flavour, or architecture.
+- If installing from a local `.lgx` file: Basecamp reports an error if the archive has no variant for your platform.
+
+Either way, confirm the archive includes a variant matching your platform (`linux-x86_64`, `linux-aarch64`, `darwin-x86_64`, or `darwin-arm64`) before reinstalling. If the package shows as installed in Package Manager but still doesn't appear in the relevant inspector, its manifest `type` may not match where you're looking—core modules only appear in Module Inspector, and UI modules (`ui_qml`) only appear in Apps Inspector.
```
weboko: These aren't the names that appear *inside* an `.lgx` archive, and this PR's own new reference page uses the other set.

`docs/core/build-modules/build-and-run-a-logos-core-module.md` calls out the distinction directly:

> `logosctl` error messages report the platform as `linux-x86_64` while LGX manifests label it `linux-amd64`. These refer to the same architecture.

So `linux-x86_64` / `linux-aarch64` are what the *tooling prints*, whereas this sentence asks the reader to "confirm the archive includes a variant matching your platform" — i.e. the manifest/variant directory names. Everywhere else in the repo those are `linux-amd64` / `linux-arm64` / `darwin-arm64`, including `docs/core/reference/lgx-package-format-and-bundling-reference.md` added in this same PR (`variants/linux-amd64/`, `variants/darwin-arm64/`).

`darwin-x86_64` doesn't appear anywhere else in the repo at all, and the swap page states the opposite — that Intel macOS is precisely the platform with no matching build.

```suggestion
Either way, confirm the archive includes a variant matching your platform (`linux-amd64`, `linux-arm64`, or `darwin-arm64`) before reinstalling. Note that the catalogue publishes no Intel macOS build. If the package shows as installed in Package Manager but still doesn't appear in the relevant inspector, its manifest `type` may not match where you're looking—core modules only appear in Module Inspector, and UI modules (`ui_qml`) only appear in Apps Inspector.
```

The same four names appear in the prerequisites at line 33 (not a line this PR touches, so CI won't flag it) — worth fixing in the same pass so the page is self-consistent.

docs/core/reference/lgx-package-format-and-bundling-reference.md
```diff
@@ -11,4 +11,63 @@ slug: lgx-package-format-and-bundling-reference
...
+├── manifest.sig           # Optional - Ed25519 signature with DID identity
+├── assets/                # Optional - variant-independent package assets
+│   ├── icon.png           #   Package icon: PNG, exactly 256x256
+│   └── lidl/               #   Canonical module interface documents
+│       └── <name>.lidl
+├── variants/               # Required - one directory per platform build
+│   ├── linux-amd64/
+│   └── darwin-arm64/
+├── docs/                   # Optional - documentation
+└── licenses/                # Optional - license files
```
weboko: The `#` comments don't line up, which is visible in a fixed-width code block. Measured column of each `#`: 27, 27, 27, 27, **28, 28, 28, 29**.

```suggestion
├── manifest.json          # Required - package metadata
├── manifest.sig           # Optional - Ed25519 signature with DID identity
├── assets/                # Optional - variant-independent package assets
│   ├── icon.png           #   Package icon: PNG, exactly 256x256
│   └── lidl/              #   Canonical module interface documents
│       └── <name>.lidl
├── variants/              # Required - one directory per platform build
│   ├── linux-amd64/
│   └── darwin-arm64/
├── docs/                  # Optional - documentation
└── licenses/              # Optional - license files
```

docs/core/reference/lgx-package-format-and-bundling-reference.md
```diff
@@ -11,4 +11,63 @@ slug: lgx-package-format-and-bundling-reference
...
+| `icon` | string | Path to the package's `assets/icon.png`. Required for `type: "ui_qml"` |
+
+See the [full field reference](https://github.com/logos-co/logos-package/blob/master/docs/spec.md#manifest-schema) for the complete schema, including dependency version ranges and signer pins.
+
+### The `ui_qml` contract
+
+A package's `type` distinguishes a [core module](../../get-started/glossary.md#core-module) from a [UI module](../../get-started/glossary.md#ui-module):
+
+- A **core module** package uses `main` as its per-variant entry point, same as most other package types.
+- A **UI module** package sets `type: "ui_qml"` and adds a `view` field: the relative path to its QML entry point, identical across variants. `main` becomes optional for this type—when present, it points at a per-variant backend Qt plugin that the host runs in an isolated process and bridges to the QML view; when absent, the QML view loads directly in-process. `type: "ui_qml"` packages must also ship a 256x256 `assets/icon.png`, since these are the packages rendered as tiles in Basecamp's app grid and sidebar.
```
weboko: This bullet and the structure diagram contradict each other for the pure-QML case.

Here: `main` is optional for `type: "ui_qml"`, and "when absent, the QML view loads directly in-process". If `main` is absent there is no per-variant binary to ship — but the diagram above marks `variants/` as **Required - one directory per platform build**, and the prose under it says "files directly under `variants/` are not allowed".

So what does a `ui_qml` package with no `main` actually put in `variants/`? Either `variants/` is not unconditionally required, or such a package still needs a (possibly empty) variant directory per platform — and a reader building their first UI module can't tell which from this page.

Worth pinning down against the spec and stating it explicitly, e.g.:

```suggestion
- A **UI module** package sets `type: "ui_qml"` and adds a `view` field: the relative path to its QML entry point, identical across variants. `main` becomes optional for this type—when present, it points at a per-variant backend Qt plugin that the host runs in an isolated process and bridges to the QML view; when absent, the QML view loads directly in-process and the package still declares one `variants/` directory per platform it supports, even though those directories carry no binary. `type: "ui_qml"` packages must also ship a 256x256 `assets/icon.png`, since these are the packages rendered as tiles in Basecamp's app grid and sidebar.
```

Please correct the wording if the real behaviour is that `variants/` may be omitted entirely — I couldn't verify it, see the review summary.

docs/messaging/concepts/about-content-topics.md
```diff
@@ -13,6 +13,10 @@ sidebar_position: 2
 # About content topics
  
 #### Understand how content topics filter and route messages, and how naming choices affect privacy.
+
+:::info
+This document reflects the state of this Logos component as it will exist on mainnet. Some features described here may not be available on the current testnet.
+:::
```
weboko: Flagging here because it's the only line of this file in the diff.

This PR renames "dApp" to "App" in the Zone SDK tutorial (title, heading and slug) and in `introduction-to-logos.md` ("At the top of the stack sit the **apps**"). Four occurrences survive, in two files this PR already edits:

- `docs/messaging/concepts/about-content-topics.md:39` — "your decentralised application (DApp) to prevent conflicts with other DApps"
- `docs/messaging/concepts/about-content-topics.md:44` — "if your DApp is called `SuperCrypto`"
- `docs/messaging/concepts/about-content-topics.md:50` — "optimal performance of your DApp"
- `docs/messaging/concepts/understand-logos-delivery-protocols.md:50` — "like when a DApp starts"

Since the point of the audit is that "dApp" is no longer the term, leaving them behind makes the terminology inconsistent within a single product section — and the capitalisation (`DApp`) doesn't match the old convention either. Suggested replacements:

- line 39: `` `application-name`: This is the unique name of your decentralised app to prevent conflicts with other apps. ``
- line 44: `For example, if your app is called `SuperCrypto` and it allows users to...`
- line 50: `...ensures optimal performance of your app.`
- `understand-logos-delivery-protocols.md:50`: `...like when an app starts.`

I couldn't attach these as suggestions — those lines aren't in this PR's diff.

.github/styles/config/vocabularies/Logos/accept.txt
```diff
@@ -134,6 +139,7 @@ lookups
 mainnet
 Mantle
 marker's
+Merkle
 MDX
 mempool
 microkernel
```
weboko: Nit: the file is sorted case-insensitively, and `Merkle` landed before `MDX`. It belongs after `mempool`.

```suggestion
marker's
MDX
mempool
Merkle
microkernel
```

(No functional impact — Vale doesn't care about order — but it keeps the list greppable.)

weboko (commented):
Reviewed and dogfooded this branch end to end (`cf9f48b`) on a clean Linux container.

## What I ran

| Check | Result |
|---|---|
| `npm ci && npm run build` | ✅ passes, no broken-link errors |
| `vale --minAlertLevel=error` on all 24 changed `.md` files (v3.11.1, matching CI) | ✅ no errors on added lines |
| Every new/changed relative link and glossary anchor (`#package`, `#module`, `#catalogue`, `#core-module`, `#ui-module`) | ✅ all resolve |
| Removed assets + removed `### Module repo` anchor still referenced anywhere? | ✅ no dangling references |
| The two new `computationally infeasible` citations to `lip.logos.co` | ✅ page and deep anchor both resolve |

So CI should be green. The substance is where the issues are.

## Blocking

**The `dApp` → `App` rename changes a live public URL with no redirect.** `/build-an-app/tutorials/build-dapp-with-logos-zone-sdk` serves a real page on `docs.logos.co` right now and 404s after this merges. Details and the exact `docusaurus.config.ts` snippet are in the inline comment. This is the one thing I'd hold the merge on — `AGENTS.md` calls it out specifically because some of those URLs are spoken in demo videos and embedded in in-app feedback links, where they can't be fixed after the fact.

## Should fix

- **Platform variant names in the Basecamp page contradict the new LGX reference page added in this same PR** (`linux-x86_64`/`darwin-x86_64` vs `linux-amd64`/`darwin-arm64`). `darwin-x86_64` appears nowhere else in the repo, and the swap page says Intel macOS is exactly the platform with *no* build. Inline, with a suggestion.
- **Four `DApp` occurrences survive the rename**, in two files this PR already edits. Inline, with replacements.
- **`variants/` is marked Required, but `main` is optional for `ui_qml`** — the new reference page contradicts itself on what a pure-QML package ships. Inline.

## Nits

- Code-fence comment alignment in the LGX structure diagram (suggestion inline).
- `Merkle` is out of alphabetical order in `accept.txt` (suggestion inline).
- `introduction-to-the-logos-blockchain.md:17` — the new sentence breaks the verb parallelism it used to have: "It **is designed to be used** for building …, **gives** those applications …, and **allows** them to …". The old "It hosts …, gives …, and guarantees …" read cleanly. Purely stylistic, but the legal rewrite is what introduced it.
- Pre-existing typo on a page this PR touches: `faq-logos-blockchain.md:49` has `decentalisation` (missing `r`). Vale flags it at error level but `filter_mode: added` means CI won't fail — worth sweeping up while you're in the file.

## Question on banner coverage

The mainnet/testnet `:::info` banner landed on all 19 concept, intro and FAQ pages **except** the five under `docs/messaging/concepts/understand-peer-discovery/` (`about-discv5.md`, `about-dns-discovery.md`, `about-peer-exchange.md`, `about-static-peers.md`, `understand-peer-discovery.md`). Every sibling in `docs/messaging/concepts/` got it, so this looks like a glob that didn't recurse rather than a decision. Intentional?

## What I could not verify — and it's the riskiest part of this PR

Two areas need a human with access to the source repos, because GitHub access for this session is scoped to `logos-docs` only:

1. **The Basecamp UI rewrite.** This PR moves the whole navigation model: loading a module goes from a **Modules** view in the sidebar to **Settings → Module Inspector / Apps Inspector**, plus new strings (**Install Local Package**, the **All / Installed / Not Installed** tabs, the **Type** column, the **Not available** row state). I could not download the AppImage or read the QML to confirm any of it. `AGENTS.md` is clear that the app's own source at the release tag is the authority here and that no CI can detect drift — please have someone check these against `logos-co/logos-basecamp` at the Testnet v0.2.1 tag before merging. If they're right this is a genuinely better page; if they're wrong it's unusable, and nothing in this repo will tell you which.

2. **The new LGX reference page's technical claims** — deterministic builds, the Merkle content-hash tree, the Ed25519/`did:jwk` signing flow, `--require-signatures`, `trustedSigners` in `logos-repo.json`, and the manifest field table. These come from `logos-co/logos-package`, which I couldn't read. The two spec links returned `403` through this environment's egress proxy rather than a clean `200`/`404`, so I can't even confirm the links resolve — worth a manual click.

Everything else above I verified directly.

docs/core/reference/lgx-package-format-and-bundling-reference.md
kashepavadan: This appears to be a problem also in the original doc. @dlipicar can you take a look please?

docs/core/reference/lgx-package-format-and-bundling-reference.md
```diff
@@ -11,4 +11,63 @@ slug: lgx-package-format-and-bundling-reference
 
 # LGX package format and bundling reference
 
-We're working on the content.
+An LGX file (`.lgx`) is the [package](../../get-started/glossary.md#package) format for distributing a Logos [module](../../get-started/glossary.md#module): a deterministic, gzip-compressed tar archive that bundles one or more platform-specific builds of the module with a signed manifest describing its metadata and dependencies. `lgx` is the CLI that creates, inspects, and signs these archives; `lgpm` and `lgpd` install them on a running system or from a [catalogue](../../get-started/glossary.md#catalogue).
+
+This page summarises the format. The [`logos-package` spec](https://github.com/logos-co/logos-package/blob/master/docs/spec.md) is the authoritative reference for the manifest schema, CLI commands, and signing details; the [catalogue format spec](https://github.com/logos-co/logos-modules-release-tool/blob/master/docs/catalog-format.md) covers how packages are listed and verified for download.
```
weboko: `2116edf` changed the catalogue-spec link in **Further reading** (line 72) from `/blob/master/` to `/blob/main/`, but this line points at the *same file in the same repo* and still says `master`. One page, two different refs for one document.

I checked which is actually right:

```
$ curl -o /dev/null -w "%{http_code}\n" -L \
  https://raw.githubusercontent.com/logos-co/logos-modules-release-tool/main/docs/catalog-format.md
200
$ curl -o /dev/null -w "%{http_code}\n" -L \
  https://raw.githubusercontent.com/logos-co/logos-modules-release-tool/master/docs/catalog-format.md
200
$ diff cat-main.md cat-master.md && echo IDENTICAL
IDENTICAL
```

Both resolve and are byte-identical, so this is consistency rather than a broken link. Matching line 72 and using `main`:

```suggestion
This page summarises the format. The [`logos-package` spec](https://github.com/logos-co/logos-package/blob/master/docs/spec.md) is the authoritative reference for the manifest schema, CLI commands, and signing details; the [catalogue format spec](https://github.com/logos-co/logos-modules-release-tool/blob/main/docs/catalog-format.md) covers how packages are listed and verified for download.
```

Worth noting the `logos-package` link in the same sentence must **stay** on `master` — that repo has no `main`:

```
$ curl -o /dev/null -w "%{http_code}\n" -L \
  https://raw.githubusercontent.com/logos-co/logos-package/main/docs/spec.md
404
```

So the two links legitimately differ, which is exactly why it reads like a typo. Leaving both on their correct-but-different refs is fine; having one file referenced two ways on one page is not.

docs/core/reference/lgx-package-format-and-bundling-reference.md
```diff
@@ -11,4 +11,63 @@ slug: lgx-package-format-and-bundling-reference
...
+
+Building a package with `lgx` is deterministic: identical inputs always produce a byte-identical `.lgx`, which lets a catalogue verify a download against a published content hash rather than trusting the transport.
+
+## Manifest fields
+
+`manifest.json` is a UTF-8 JSON file. The fields most relevant to authoring and consuming a package:
+
+| Field | Type | Purpose |
+|---|---|---|
+| `name`, `version` | string | Canonical lowercase package identity |
```
weboko: The table omits `manifestVersion`, which is a **required** field — and it is the field that gates two of the rules this same page describes.

From the upstream spec: the icon contract applies at `0.4.0`+, and `optional_dependencies` (listed two rows below) is `0.5.0`/`0.6.0`+. A reader authoring a manifest from this table has no way to know either of those is version-dependent, and no way to know what to put in the one field that decides it.

I confirmed the gating is real, not just documented — same package, only `manifestVersion` changed:

```
$ lgx verify noicon060.lgx      # ui_qml, no icon, manifestVersion 0.6.0
Error: Package validation failed:
  - Missing 'icon': assets/icon.png is required for type 'ui_qml' packages

$ lgx verify noicon030.lgx      # ui_qml, no icon, manifestVersion 0.3.0
Package structure is valid: noicon030.lgx
```

```suggestion
| `manifestVersion` | string | Manifest schema version, currently `0.6.0`. Gates the version-dependent rules on this page: the icon contract (`0.4.0`+) and `optional_dependencies` (`0.6.0`+) |
| `name`, `version` | string | Canonical lowercase package identity |
```

docs/core/reference/lgx-package-format-and-bundling-reference.md
```diff
@@ -11,4 +11,63 @@ slug: lgx-package-format-and-bundling-reference
...
+| `type` | string | Package classification, for example `core`, `library`, or `ui_qml` |
+| `dependencies` | array | Other packages this one requires to run |
+| `optional_dependencies` | array | Packages this one can call but does not require |
+| `main` | object | Map of platform variant name to the entry point path for that variant |
+| `display_name` | string | Human-readable label shown by UI consumers (Package Manager, App Manager) and CLI tools. Falls back to `name` when absent |
```
weboko: **App Manager** is not a name this documentation uses anywhere else. It is lifted verbatim from the upstream spec, but in this repo — including `install-and-load-a-module-in-logos-basecamp.md`, which *this PR edits* — the Basecamp surface is consistently called **Apps Inspector**:

```
$ grep -rn "App Manager\|Apps Inspector" docs/
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:38:  ... core modules under **Module Inspector**, UI modules under **Apps Inspector**.
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:39:  **Apps Inspector** shows a UI module's status and version ...
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:66:  ... or **Apps Inspector** for a UI module.
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:75:  ### The installed module doesn't appear in Module Inspector or Apps Inspector
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:82:  ... only appear in Apps Inspector.
docs/core/reference/lgx-package-format-and-bundling-reference.md:50:  ... (Package Manager, App Manager) ...
```

A reader who goes looking for "App Manager" in Basecamp will not find it. Since a terminology audit is the whole point of this PR, this is the kind of thing worth catching:

```suggestion
| `display_name` | string | Human-readable label shown by UI consumers (Package Manager, Apps Inspector) and CLI tools. Falls back to `name` when absent |
```

If the upstream "App Manager" is genuinely a *different* surface from Basecamp's Apps Inspector, then this row needs a word distinguishing them rather than my rename — but as written the two read as the same thing under two names.

docs/core/reference/lgx-package-format-and-bundling-reference.md
```diff
@@ -11,4 +11,63 @@ slug: lgx-package-format-and-bundling-reference
...
+| `provides` | array | App-to-app intents the package can service, for example `chat.group.open` |
+| `icon` | string | Path to the package's `assets/icon.png`. Required for `type: "ui_qml"` |
```
weboko: "Required for `type: "ui_qml"`" is unconditional here, but the contract only applies from `manifestVersion` `0.4.0` onward. The spec is explicit that older manifests are exempt, because `icon: ""` was legal at `0.3.0` and applying the rule retroactively would make every already-published package fail verification.

`lgx --help` states the gate too:

```
--icon <path>   Package icon: PNG, exactly 256x256. Stored
                once at assets/icon.png and shared by every
                variant. Required for `ui_qml` packages at
                manifest version 0.4.0+
```

And it is enforced exactly that way (verified above on line 45).

```suggestion
| `icon` | string | Path to the package's `assets/icon.png`. Required for `type: "ui_qml"` at `manifestVersion` `0.4.0`+; optional for every other type |
```

One detail worth knowing while you are here: the size rule is enforced whenever an icon is *present*, whatever the `type`, and only at `lgx verify` / `lgx sign` / `lgpm install` — `lgx add --icon` silently accepted 128x128, 512x512 and 256x255 for me and only `verify` rejected them. That is intended per the spec ("a package may be assembled in any order"), so no change needed, but it means a bundling pipeline that never calls `verify` will ship a bad icon.

docs/core/reference/lgx-package-format-and-bundling-reference.md
```diff
@@ -11,4 +11,63 @@ slug: lgx-package-format-and-bundling-reference
...
+
+See the [full field reference](https://github.com/logos-co/logos-package/blob/master/docs/spec.md#manifest-schema) for the complete schema, including dependency version ranges and signer pins.
+
+### The `ui_qml` contract
+
+A package's `type` distinguishes a [core module](../../get-started/glossary.md#core-module) from a [UI module](../../get-started/glossary.md#ui-module):
+
+- A **core module** package uses `main` as its per-variant entry point, same as most other package types.
+- A **UI module** package sets `type: "ui_qml"` and adds a `view` field: the relative path to its QML entry point, identical across variants. `main` becomes optional for this type—when present, it points at a per-variant backend Qt plugin that the host runs in an isolated process and bridges to the QML view; when absent, the QML view loads directly in-process. `type: "ui_qml"` packages must also ship a 256x256 `assets/icon.png`, since these are the packages rendered as tiles in Basecamp's app grid and sidebar.
```
weboko: @kashepavadan @dlipicar — this answers the thread on this bullet, so I do not think it needs to block.

The apparent contradiction ("`variants/` Required" vs "`main` optional") dissolves once you separate the directory from its contents. `variants/` is required as a **directory**; it is not required to contain any variant. `lgx create` emits it empty and the package verifies:

```
$ lgx create calcui && tar tzvf calcui.lgx
-rw-r--r-- 0/0   193  manifest.json
drwxr-xr-x 0/0     0  variants/
$ lgx verify calcui.lgx
Package structure is valid: calcui.lgx
```

And a QML-only package needs no `main` at all — the variant directory holds the QML and nothing else:

```
$ lgx add calcui4.lgx --variant linux-amd64 --files ./src/Main.qml --view Main.qml --icon icon.png
Added variant 'linux-amd64' to calcui4.lgx
$ tar tzf calcui4.lgx
assets/icon.png
manifest.json
variants/linux-amd64/Main.qml
$ lgx verify calcui4.lgx
Package structure is valid
```

`lgx add --help` says the same: `--main` is "required for directory variants except `ui_qml`, where `main` is optional backend metadata".

Folding that in, plus the `0.4.0` icon gate from the row above:

```suggestion
- A **UI module** package sets `type: "ui_qml"` and adds a `view` field: the relative path to its QML entry point, identical across variants. `main` becomes optional for this type—when present, it points at a per-variant backend Qt plugin that the host runs in an isolated process and bridges to the QML view; when absent, the QML view loads directly in-process and the package's variant directories carry only its QML files. `variants/` is required either way: it is the directory that is mandatory, not its contents, so a package with no native binary is still valid. From `manifestVersion` `0.4.0` onward, `type: "ui_qml"` packages must also ship a 256x256 `assets/icon.png`, since these are the packages rendered as tiles in Basecamp's app grid and sidebar.
```

weboko (commented):
Re-reviewed and re-dogfooded at `5127c83`, covering the four commits pushed since my last review (`bc90f73`, `2116edf`, `a70f228`, `5127c83`).

## The blocker is fixed

The renamed-slug redirect is in, and I verified it against a real build rather than just reading the diff:

```
$ npm ci && npm run build
[SUCCESS] Generated static files in "build".

$ ls build/build-an-app/tutorials/
build-app-with-logos-zone-sdk
build-dapp-with-logos-zone-sdk     # <- redirect stub, was missing before

$ cat build/build-an-app/tutorials/build-dapp-with-logos-zone-sdk/index.html
<meta http-equiv="refresh" content="0; url=/build-an-app/tutorials/build-app-with-logos-zone-sdk">
window.location.href = '...' + window.location.search + window.location.hash;
```

Query string and hash are carried across, so deep links survive. Nothing blocking left from my side.

Also confirmed fixed: the four surviving `DApp` occurrences, and the code-fence comment alignment (all `#` now at column 27).

## CI status

| Check | Result |
|---|---|
| `npm ci && npm run build` | passes, no broken-link errors |
| `vale --minAlertLevel=error` on all 24 changed `.md` files (v3.11.1, matching CI) | 1 error, on a **pre-existing** line |

The one Vale error is still `faq-logos-blockchain.md:49` `decentalisation`. It is not on an added line, so `filter_mode: added` keeps CI green — but it is in a file this PR already touches.

## This time I could verify the LGX page against the real tooling

Last time I flagged that I could not check the upstream claims. I can now: I built `lgx` from `logos-co/logos-package@master` and exercised the format.

<details>
<summary>Build + test transcript</summary>

```
$ apt install cmake libicu-dev zlib1g-dev libsodium-dev
$ cmake .. && make -j$(nproc)
[100%] Built target lgx
$ ./lgx --version
lgx version 0.1.0
```

**Deterministic build** — two independent builds of identical inputs, 2s apart:
```
735d71f6316d1d7b0aa89a04d352e81992ea0544c7e17266c31f6c109eb9b0d8  r1/same.lgx
735d71f6316d1d7b0aa89a04d352e81992ea0544c7e17266c31f6c109eb9b0d8  r2/same.lgx
BYTE-IDENTICAL
```

**Root entry constraints:**
```
$ lgx verify bad3.lgx
Error: Package validation failed:
  - Forbidden root entry: evil

$ lgx verify bad2.lgx
Error: Package validation failed:
  - File directly under variants/: variants/stray.txt
```

**Icon exactly 256x256:**
```
$ lgx verify t128.lgx
Error: Icon 'assets/icon.png' does not match the Logos icon standard:
       expected PNG, exactly 256x256; actual: PNG, 128x128
```
</details>

So the page's **deterministic build**, **root-entry**, **files-directly-under-`variants/`** and **exactly-256x256** claims all check out against the real tool. Three things did not, and they are inline below.

## The open `variants/` question is answerable — no need to block on it

@kashepavadan asked @dlipicar about the `variants/` Required vs `main` optional contradiction. The tooling answers it directly:

```
$ lgx create calcui
$ tar tzvf calcui.lgx
-rw-r--r-- 0/0   193  manifest.json
drwxr-xr-x 0/0     0  variants/          <- present, empty

$ lgx verify calcui.lgx
Package structure is valid: calcui.lgx
```

`variants/` is required as a **directory**; it is not required to contain anything. `lgx create` emits it empty and `lgx verify` passes. And a QML-only package builds fine with no `--main` at all:

```
$ lgx add calcui4.lgx --variant linux-amd64 --files ./src/Main.qml --view Main.qml --icon icon.png
Added variant 'linux-amd64' to calcui4.lgx
$ tar tzf calcui4.lgx
assets/icon.png
manifest.json
variants/linux-amd64/Main.qml    <- variant dir, QML only, no binary
$ lgx verify calcui4.lgx
Package structure is valid
```

So the resolution is: the directory is mandatory, its contents are not, and a pure-QML package ships variant directories carrying only QML. Suggested wording inline on the bullet.

## Still open from the last round

- `Merkle` is still out of alphabetical order in `accept.txt` (after `marker's`, before `MDX`; belongs after `mempool`). Cosmetic, nothing depends on it.
- The banner-coverage question is still unanswered: the five pages under `docs/messaging/concepts/understand-peer-discovery/` are the only concept pages that did not get the mainnet/testnet `:::info` banner. Still looks like a glob that did not recurse. Not blocking — just needs a yes/no.

## Still not verifiable here

The Basecamp UI rewrite (**Settings → Module Inspector / Apps Inspector**, **Install Local Package**, the **All / Installed / Not Installed** tabs, the **Type** column, the **Not available** row state) is unchanged since my last review and still needs someone with the app to check it against `logos-co/logos-basecamp` at the Testnet v0.2.1 tag. `AGENTS.md` is explicit that no CI can detect drift here.

Dropping the platform-variant list from both spots resolves the `darwin-x86_64` contradiction I raised, so that one is closed. It does leave "confirm the archive includes a variant matching your platform" with no indication of what a variant name looks like — the LGX reference page added in this PR has the canonical names, so a link from that sentence would close the loop. Non-blocking.

docs/core/reference/lgx-package-format-and-bundling-reference.md
```diff
@@ -11,4 +11,64 @@ slug: lgx-package-format-and-bundling-reference
...
+
+### The `ui_qml` contract
+
+A package's `type` distinguishes a [core module](../../get-started/glossary.md#core-module) from a [UI module](../../get-started/glossary.md#ui-module):
+
+- A **core module** package uses `main` as its per-variant entry point, same as most other package types.
+- A **UI module** package sets `type: "ui_qml"` and adds a `view` field: the relative path to its QML entry point, identical across variants. `main` becomes optional for this type—when present, it points at a per-variant backend Qt plugin that the host runs in an isolated process and bridges to the QML view; when absent, the QML view loads directly in-process and the package's variant directories carry only its QML files. `variants/` is required either way: it is the directory that is mandatory, not its contents, so a package with no native binary is still valid. From `manifestVersion` `0.4.0` onward, `type: "ui_qml"` packages must also ship a 256x256 `assets/icon.png`, since these are the packages rendered as tiles in Basecamp's app grid and sidebar.
```
weboko: **This is why `vale` / `prose` are red on `59a1c56`.** `Logos.SpellingGB` is a plain en_GB dictionary, and it rejects `onward` — British English wants the adverbial `onwards`. The alert is on an added line, so `filter_mode: added` + `fail_on_error: true` makes it blocking.

Reproduced locally with the same Vale version CI pins (3.11.1):

```
$ vale --minAlertLevel=error docs/core/reference/lgx-package-format-and-bundling-reference.md
 62:602  error  'onward' is not in British English.  Logos.SpellingGB
1 error, 0 warnings and 0 suggestions in 1 file.
```

And with `onwards`:

```
0 errors, 0 warnings and 0 suggestions in 1 file.
```

`onwards` is the better fix than an `accept.txt` entry here — `onward` is a real word, just the American adverb form, so allowlisting it would let the wrong form through everywhere else too.

```suggestion
- A **UI module** package sets `type: "ui_qml"` and adds a `view` field: the relative path to its QML entry point, identical across variants. `main` becomes optional for this type—when present, it points at a per-variant backend Qt plugin that the host runs in an isolated process and bridges to the QML view; when absent, the QML view loads directly in-process and the package's variant directories carry only its QML files. `variants/` is required either way: it is the directory that is mandatory, not its contents, so a package with no native binary is still valid. From `manifestVersion` `0.4.0` onwards, `type: "ui_qml"` packages must also ship a 256x256 `assets/icon.png`, since these are the packages rendered as tiles in Basecamp's app grid and sidebar.
```

weboko (commented):
### Re-review + dogfooding run on `59a1c56`

Re-ran this against the current head because the batched-suggestions commit landed after my last pass. **One blocking item**, and it is the reason CI is red.

#### Verified fixed since my last review

- `manifestVersion` row added to the manifest table, and the icon contract is now correctly gated on `0.4.0`+. 
- Catalogue-spec link is consistent at `/blob/main/` in both places. I checked both repos' actual default branches — `logos-co/logos-package` is `master`, `logos-co/logos-modules-release-tool` is `main` — so every ref on the page is now correct, and both spec URLs return `200`.
- Structure-diagram `#` comments now line up: measured column 28 for all eight.
- **The renamed page's URL is safe.** The `client-redirects` entry landed and I confirmed it against a real build, including that deep links survive:
  ```
  $ npm run build && cat build/build-an-app/tutorials/build-dapp-with-logos-zone-sdk/index.html
  <meta http-equiv="refresh" content="0; url=/build-an-app/tutorials/build-app-with-logos-zone-sdk">
  window.location.href = '/build-an-app/tutorials/build-app-with-logos-zone-sdk'
                         + window.location.search + window.location.hash
  ```
  `sidebars.ts` points at the real route, not the stub. 
- No `dApp`/`DApp`/`dapp` occurrences left anywhere under `docs/`. 
- The `linux-x86_64` vs `linux-amd64` mismatch is gone (the platform list was dropped rather than corrected — that works).
- Removing the **Module repo** glossary entry is safe: nothing in the repo links to `glossary.md#module-repo`.

#### Dogfooding: `build-app-with-logos-zone-sdk`

This is the one runnable doc the PR touches, so I followed it end to end as a reader on a clean machine, rather than only reading the rename.

Every path the page cites exists on the `tutorial` branch (`sequencer/src/sequencer.rs`, `indexer/src/indexer.rs`, `common/src/state.rs`, `sequencer/src/db.rs`, both zone-sdk paths, and the root `README.md`). I then extracted all 15 Rust snippets from the page, merged the three elided `impl Sequencer` / two `impl Indexer` blocks as the page instructs, dropped them into the three `// Your code here` markers, and built:

```
$ cargo build
   Compiling demo-sqlite-common v0.2.1
   Compiling demo-sqlite-sequencer v0.2.1
   Compiling demo-sqlite-indexer v0.2.1
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 15s
```

**The tutorial's code compiles as written** — the "compilable password manager Zone demo" claim in *What to expect* holds, and the `new` function matches the reference implementation on `master` line for line. 

Two build notes for anyone else reproducing this, neither a docs problem:
- The prerequisites list Rust stable, but the repo pins `1.94.0` via `rust-toolchain.toml`, so that resolves itself.
- `logos-blockchain-circuits-pol-sys`'s build script downloads a prebuilt artifact with `ureq`, which uses its own baked-in root store. Behind a TLS-inspecting proxy that fails with `invalid peer certificate: UnknownIssuer`. Pre-seeding `~/.cache/logos/blockchain/` (or setting `LBC_ROOT_DIR`) works around it.

#### Non-blocking, pre-existing — flagging, not asking you to fix here

These are in the tutorial's body, which this PR only renames, so they are out of scope. Recording them so they are not lost:

- The tutorial's `handle_event` binds `Event::BlocksProcessed { checkpoint, channel_update, finalized }`, which makes every reader who follows the page get `warning: unused variable: channel_update` on their first build. The page's own prose says a centralised Zone ignores it, and `master` writes `channel_update: _` — so the doc diverges from the implementation it is teaching.
- In the `common/src/state.rs` snippet, the four leading `//` comment lines sit at 3-space indent while the code below them sits at 4, so the rendered block shows every line of actual code offset by one space. Every other Rust block on the page is uniformly at 3.
- `docs/blockchain/faq-logos-blockchain.md:49` has `decentalisation` (missing `r`). Vale flags it at error level, but on a pre-existing line, so `filter_mode: added` correctly lets it through and it does not block you. Cheap to sweep up while you are in the file.
- The `Merkle` sort-order thread got resolved without the change being applied — it still sits between `marker's` and `MDX` in `accept.txt`. Purely cosmetic, Vale does not care about order.

#### Still open

The **App Manager** thread on line 51 is unresolved and still reproduces at this head — replied there with the current `grep`.

docs/core/reference/lgx-package-format-and-bundling-reference.md
weboko: Still reproduces at `59a1c56` — re-checked on the current head:

```
$ grep -rn "App Manager\|Apps Inspector\|Module Inspector" docs/
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:38:  ... core modules under **Module Inspector**, UI modules under **Apps Inspector**.
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:39:  **Module Inspector** also shows ... **Apps Inspector** shows a UI module's status and version ...
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:66:  Open **Module Inspector** for a core module, or **Apps Inspector** for a UI module.
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:75:  ### The installed module doesn't appear in Module Inspector or Apps Inspector
docs/basecamp/install-and-load-a-module-in-logos-basecamp.md:82:  ... core modules only appear in Module Inspector, and UI modules (`ui_qml`) only appear in Apps Inspector.
docs/core/reference/lgx-package-format-and-bundling-reference.md:51:  ... shown by UI consumers (Package Manager, App Manager) ...
```

Six occurrences, and this line is the only one that says **App Manager**. The other five are all in `install-and-load-a-module-in-logos-basecamp.md`, which this PR rewrites specifically to introduce **Apps Inspector** as the name for that surface. So the PR teaches a reader one name on one page and a different name for the same thing on another.

```suggestion
| `display_name` | string | Human-readable label shown by UI consumers (Package Manager, Apps Inspector) and CLI tools. Falls back to `name` when absent |
```

If **App Manager** is deliberate because it is the upstream spec's term rather than Basecamp's UI label, then it needs saying explicitly — otherwise a reader looking for "App Manager" in Basecamp will not find it.

